PR #40, opened 2017-04-26
feat(tracks): add tracks framework

yairans:


kaltura-hooks:
Hi @yairans,

itaykinnrot:
@yairans in general i think that id3 tags also count as tracks no? - metadata

src/engines/adapters/base-adapter.js
```diff
@@ -113,4 +115,39 @@ export default class BaseMediaSourceAdapter {
   destroy() {
     // should do nothing. implemented by the inheritor if necessary.
   }
+
+  selectTrack(track: Track) {
+    if (track) {
+      let success = false;
+      switch (track.type) {
```
dan-ziv: I don't think we need the _type_ property in Track class.
We will conclude the type by the instance of the track.
For example:
`if (track instanceof AudioTrack){...}`

src/engines/adapters/native-adapter.js
```diff
@@ -50,4 +53,34 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
   load(): void {
     this._msPlayer.load();
   }
+
+  _selectAudioTrack(track: AudioTrack): boolean {
+    if (track && track.type === TrackTypes.AUDIO && this._msPlayer.audioTracks && this._msPlayer.audioTracks.length) {
```
dan-ziv: Why to do a recheck of the type if we already did it in _selectTrack_?

src/track/audio-track.js
```diff
@@ -0,0 +1,7 @@
+//@flow
+import Track from './track'
+import TrackTypes from './track-types'
+
+export default class AudioTrack extends Track {
+  _type: string = TrackTypes.AUDIO;
```
dan-ziv: No need this property

src/track/track.js
```diff
@@ -0,0 +1,33 @@
...
+
+  get type(): string {
+    return this._type;
+  }
+
+  get language(): string {
+    return this._language;
+  }
+
+  constructor(id: string, active: boolean, language: string) {
```
dan-ziv: I think it's generally better to send a _details_ object that will contain all the data.
In this way we don't need to care about the order and the amount of the params .
For example:
`constructor(details: Object) {`
   ` this._id = details.id;`
   ` this._active = details.active;`
   ` this._name = details.name;`
`  }`

src/track/track.js
```diff
@@ -0,0 +1,33 @@
+//@flow
+export default class Track {
+  _id: string;
+  _active: boolean;
+  _type: string;
+  _language: string;
```
dan-ziv: We need to add `_name: string;` for the UI to use it.

src/track/track.js
dan-ziv: About the _language_ property, isn't it something that common only to audio and text tracks? if so, it should be in the derived AudioTrack and TextTrack classes and not here.

flow-typed/classes/track/track.js
```diff
@@ -0,0 +1,34 @@
+//@flow
+export default class Track {
```
OrenMe: flow-typed is for flow declarations only, e.g. class or interface definitions for example.
Actual concrete classes need to be part of the src/lib directory

src/player.js
```diff
@@ -107,9 +116,14 @@ class Player extends FakeEventTarget {
   //  <editor-fold desc="Playback Interface">
   /**
    * Start/resume playback
-   * @returns {void}
+   * @returns {Promise} - The play promise
    */
-  play(): void {
+  play(): Promise {
```
OrenMe: It's better not to overload the play method with our own play promise.
play is already a promise in browsers and it has annoying nature of throwing exceptions when load is called during play.
It's OK to just wait for our internal load to finish and just call play then without a promise.
Also, the interface still says play returns void...

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -109,22 +112,205 @@ export default class NativeAdapter implements IMediaSourceAdapter {
...
-   * @returns {void}
+   * @returns {Promise<Object>} - The loaded data
    */
-  load(): void {
-    NativeAdapter._logger.debug('load');
-    this._videoElement.load();
+  load(): Promise<Object> {
+    if (!this._loadPromise) {
+      this._loadPromise = new Promise((resolve, reject) => {
+        this._eventManager.listen(this._videoElement, PlayerEvents['LOADED_METADATA'], () => {
```
dan-ziv: It equals, but I think it better to use `PlayerEvents.LOADED_METADATA` syntax for consistency,

src/playkit.js
```diff
@@ -26,6 +27,9 @@ export {registerMediaSourceAdapter};
 // Export the plugin framework
 export {registerPlugin, BasePlugin};
 
+// Export the tracks classes
+export {TRACK};
```
dan-ziv: Capital letters convention point on constant vars (like the VERSION) and here it's not the case. I think it better to export it regularly...
`export {AudioTrack, TextTrack, VideoTrack....}` and so on

src/track/index.js
```diff
@@ -0,0 +1,7 @@
+import Track from './track'
+import VideoTrack from './video-track'
+import AudioTrack from './audio-track'
+import TextTrack from './text-track'
+import TRACK_TYPES from './track-types'
+
```
dan-ziv: The name of the file is too generic (although it under track folder).
Maybe we need convention to things like that for similar use in the future.
Something like 'track-export.js'.

Having saying that, I think this file is unnecessary and we just need to copy this content to playkit.js

src/player.js
```diff
@@ -10,6 +10,11 @@ import LoggerFactory from './utils/logger'
 import Html5 from './engines/html5/html5'
 import PluginManager from './plugin/plugin-manager'
 import StateManager from './state/state-manager'
+import TrackTypes from './track/track-types'
```
dan-ziv: We need to export track types as an enum via the player (like we did with the states and the events).
` get Track(): { [track: string]: string } {
    return TrackTypes;
   }		   }`

When developers would like to get the audio track for example it will look like:
`let audioTracks = player.getTracks(player.Track.AUDIO);`

src/player.js
```diff
@@ -97,6 +104,69 @@ class Player extends FakeEventTarget {
...
+   * Mark the selected track as active
+   * @function _markActiveTrack
+   * @param {Track} track - the track to mark
+   * @returns {void}
+   * @private
+   */
+  _markActiveTrack(track: Track) {
+    let type;
+    if (track instanceof VideoTrack) {
+      type = 'video';
```
dan-ziv: Use our enums for track types.
`type = TrackTypes.VIDEO`

src/track/text-track.js
```diff
@@ -0,0 +1,15 @@
+//@flow
+import Track from './track'
+
+export default class textTrack extends Track {
```
dan-ziv: Change `textTrack` to `TextTrack`

src/track/text-track.js
```diff
@@ -0,0 +1,15 @@
...
+  _kind: string;
+
+  get kind(): string {
+    return this._kind;
+  }
+
+  constructor(settings: Object = {}) {
+    super(settings);
+    this._kind = settings.kind;
```
dan-ziv: In hls subtitles has an id. Why u decided to implement text track without id propery?

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -109,22 +112,205 @@ export default class NativeAdapter implements IMediaSourceAdapter {
...
+  }
+
+  /**
+   * Select a track
+   * @function selectTrack
+   * @param {Track} track - the track to select
+   * @returns {boolean} - success
+   * @public
+   */
+  selectTrack(track: Track): boolean {
```
dan-ziv: This method should be in the player/engine level since all the adapter has the same logic for it.
The only difference between the adapters is the implementation of the specific selection - i.e `selectAudioTrack, selectVideoTrack, selectTextTrack`and therefore this should be the api that the adapter will implement and the engine will call to.

src/engines/html5/media-source/adapters/native-adapter.js
yairans: done

src/player.js
yairans: done

src/track/text-track.js
yairans: done

src/track/text-track.js
yairans: done

src/engines/html5/media-source/adapters/native-adapter.js
yairans: done

src/playkit.js
yairans: done

src/player.js
yairans: done

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -102,22 +106,190 @@ export default class NativeAdapter implements IMediaSourceAdapter {
...
-  load(): void {
-    NativeAdapter._logger.debug('load');
-    this._videoElement.load();
+  load(): Promise<Object> {
+    if (!this._loadPromise) {
+      this._loadPromise = new Promise((resolve, reject) => {
+        this._eventManager.listen(this._videoElement, PlayerEvents.LOADED_METADATA, () => {
+          this._eventManager.unlisten(this._videoElement, PlayerEvents.LOADED_METADATA);
+          let data = {tracks: this._getParsedTracks()};
+          NativeAdapter._logger.debug('load');
```
OrenMe: @yairans please be more accurate with log output, one word is not beneficial for logs clarity

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -102,22 +106,190 @@ export default class NativeAdapter implements IMediaSourceAdapter {
...
-   * Clearing the video source
+   * Clear the video source
    * @function destroy
    * @returns {void}
    */
   destroy(): void {
     NativeAdapter._logger.debug('destroy');
-    this._videoElement.src = "";
+    this._eventManager.destroy();
+    this._loadPromise = null;
```
OrenMe: @yairans don't think clearing the promise on destroy is enough, what happens if we change source but engine is the same? do we destroy and recreate Html5?

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -102,22 +106,190 @@ export default class NativeAdapter implements IMediaSourceAdapter {
...
+  /**
+   * Select a video track
+   * @function selectVideoTrack
+   * @param {VideoTrack} videoTrack - the track to select
+   * @returns {boolean} - success
+   * @public
+   */
+  selectVideoTrack(videoTrack: VideoTrack): boolean {
+    let videoTracks = this._videoElement.videoTracks;
+    if ((videoTrack instanceof VideoTrack) && videoTracks && videoTracks[videoTrack.index]) {
```
OrenMe: @yairans please verify if it is better to first disable all tracks and only then enable the new one, same for all track types.
That way you can create a method to disable tracks which just goes through all tracks and disable, and then just access the chosen track by index and set it directly as enabled.

src/engines/html5/media-source/adapters/native-adapter.js
yairans: I'm not sure about recreating Html5. but on change media the engine should call the adapter's destroy and recreate adapter according the mime type.

src/engines/html5/media-source/adapters/native-adapter.js
yairans: done

src/engines/html5/media-source/adapters/native-adapter.js
yairans: done

src/player.js
```diff
@@ -114,9 +120,8 @@ class Player extends FakeEventTarget {
     this._eventManager.destroy();
     this._pluginManager.destroy();
     this._stateManager.destroy();
-    // this.engine_ = null;
-    // this.eventManager_ = null;
-    this._config = null;
+    this._eventManager.destroy();
```
OrenMe: Already have `this._eventManager.destroy();` on line 120

PR #46, opened 2017-05-09
feat(player): handle creation of player without initial config

dan-ziv:
* Complete js-docs for the player (player.js) and the engine (html5.js).
* Add restrictions in getters and setters in case engine doesn’t exists.
* Add configure method to enable setting the configuration of the player after he already created.
* Fix flow annotations.
* Add Source type.


flow-typed/types/source.js
```diff
@@ -0,0 +1,5 @@
+declare type Source = {
```
OrenMe: No point in defining a source as an object or null.
If you intend to nullify it in your flow then use `?Source` where you hold your sources

dan-ziv: done

PR #51, opened 2017-06-05
feat(events): tracks events

yairans:


src/playkit.js
```diff
@@ -9,6 +9,10 @@ import Track from './track/track'
 import VideoTrack from './track/video-track'
 import AudioTrack from './track/audio-track'
 import TextTrack from './track/text-track'
+import FakeEventTarget from './event/fake-event-target'
+import FakeEvent from './event/fake-event'
+import {CUSTOM_EVENTS as CustomEvents} from './event/events'
+import * as TestUtils from './utils/test-utils'
```
OrenMe: Why do we need this for production?

dan-ziv: Adapters... 
Will be removed if we will use base adapter.

OrenMe: Please find other solutions for testing, we can't have testing utils leak out to our production.
test utils should be kept under tests, and not under src lib

dan-ziv:
Agree.

src/playkit.js
dan-ziv: 100%

dan-ziv: done

PR #58, opened 2017-06-13
chore(webpack): webpack local build before push

dan-ziv:


package.json
```diff
@@ -14,11 +14,24 @@
     "test:safari": "NODE_ENV=test karma start --color --browsers Safari",
     "test:watch": "NODE_ENV=test karma start --color --auto-watch",
     "start": "webpack-dev-server",
-    "release": "NODE_ENV=production npm run build -- -p && git add --all dist && git commit -m'Update dist' && standard-version",
+    "release": "npm run build:prod && npm run commit:dist && standard-version",
     "eslint": "eslint . --color",
     "flow": "flow check",
-    "prepush": "npm run eslint && npm run test && npm run flow"
+    "eslint:flow:test": "npm run eslint && npm run flow && npm run test",
+    "commit:dist": "git add . && (git commit -m 'update dist' || exit 0)",
```
OrenMe: Need to protect from adding all, I think all changes should be staged before committing the dist and this can cause additional artifacts to be added unintentionally

package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "playkit-js",
   "version": "0.1.0",
-  "main": "dist/playkit.js",
```
OrenMe: If you remove the main then any library that uses Playkit will have to use the specific path when importing

dan-ziv: I removed it by mistake, done

package.json
dan-ziv: done

PR #61, opened 2017-06-14
feat(player): report first play event

yairans:


src/player.js
```diff
@@ -276,6 +278,28 @@ class Player extends FakeEventTarget {
   }
 
   /**
+   * @function _onPlay
+   * @return {void}
+   * @private
+   */
+  _onPlay(): void {
+    this._eventManager.unlisten(this, Html5Events.PLAY);
```
OrenMe: This will cause the player to stop listening to play event altogether.
Better have a handler with a flag, so we can also query this from outside, e.g.  if playback has ever started before.

src/player.js
```diff
@@ -116,6 +111,11 @@ class Player extends FakeEventTarget {
     return {};
   }
 
+  _initializeMembers(): void {
```
OrenMe: I think in general this is not required.
The only place we initialize or reset our general members is in the constructor and in the destroy, and config and tracks can be set to default on initialization to these values.
In this case I prefer inlining.

src/player.js
```diff
@@ -276,6 +278,28 @@ class Player extends FakeEventTarget {
...
+  }
+
+  /**
+   * @function _onEnded
+   * @return {void}
+   * @private
+   */
+  _onEnded(): void {
+    this._eventManager.listen(this, Html5Events.PLAY, () => {
+      this._eventManager.unlisten(this, Html5Events.PLAY);
```
OrenMe: I think this assumption is problematic.
This means on each ended we will register the PLAY event, so even if you do change media I think this will get triggered, no?

src/player.js
yairans: Whatever you say. will be done.

src/player.js
yairans: I so hate falgs...... (player V2...)
but you are right.
we must pass an handler to unlisten to avoid  this mistake in future

src/player.js
yairans: we have anyway to clean the eventManager on change media i think

test/src/player.spec.js
```diff
@@ -446,4 +446,42 @@ describe('Track enum', function () {
...
+    });
+
+    it('should fire first play only once', (done) => {
+      let counter = 0;
+      let onPlaying = () => {
+        player.removeEventListener(Html5Events.PLAYING, onPlaying);
+        player.pause();
+        player.play();
+        setTimeout(() => {
+          counter.should.equal(1);
```
dan-ziv: I think in general it's better to use sinon spies to check if method/event has been called  and how many times.

yairans: this is a test. i prefer to test like this. i'm appreciate your review but it should't to be 'requested changes'  !!!

dan-ziv: No need to get upset :)
This is a test, but using sinon for this check is also shorter and also more robustness.

PR #62, opened 2017-06-15
ready promise

yairans:


src/player.js
```diff
@@ -286,6 +299,33 @@ class Player extends FakeEventTarget {
...
+  load(): void {
+    if (this._engine) {
+      this._engine.load().then((data) => {
+        this._tracks = data.tracks;
+        this.dispatchEvent(new FakeEvent(CustomEvents.TRACKS_CHANGED, {tracks: this._tracks}));
+      }).catch((error) => {
+        this.dispatchEvent(new FakeEvent(Html5Events.ERROR, error));
+      });
+    } else {
+      this._config.preload = 'auto';
```
OrenMe: No need to handle cases where engine is not ready due to calling player without sources

src/player.js
```diff
@@ -286,6 +299,33 @@ class Player extends FakeEventTarget {
 
   //  <editor-fold desc="Playback Interface">
   /**
+   * The player readiness
+   * @public
+   * @returns {Promise<*>} - The ready promise
+   */
+  ready(): Promise<*> {
+    return this._readyPromise ? this._readyPromise : Promise.resolve();
```
OrenMe: In which cases there's no ready promise and we resolve immediately?

src/event/events.js
```diff
@@ -94,18 +94,22 @@ const HTML5_EVENTS: { [event: string]: string } = {
...
   AUDIO_TRACK_CHANGED: 'audiotrackchanged',
   /**
-   * Fires when the text track has been changed
+   * Fires when the active text track has been changed
    */
   TEXT_TRACK_CHANGED: 'texttrackchanged',
   /**
+   * Fires when the player tracks have been changed
+   */
+  TRACKS_CHANGED: 'trackschanged',
```
dan-ziv: I think the name doesn't describe best the event meaning.
It should be
`TRACKS_READY: 'tracksready'`
or
`TRACKS_PARSED: 'tracksparsed'`

src/player.js
yairans: maybe right after changeMedia which calls destroy and null the readyPromise.

dan-ziv: in destroy create new promise which is not resolved.
`this._readyPromise = new Promise();`

src/player.js
yairans: I want to enable the user to call:
new player() -> load() -> configure(config)

OrenMe: I think load/play/pause are methods which should apply only for a given media.
Setting preload auto will persist for next media configured as well for example, which is not necessarily what a user thinks he will get if he calls load in the order you mention above.

src/event/events.js
OrenMe: @dan-ziv in high level this might be more accurate, as, for example, in live manifests the tracks can change during the live stream life time.
For example, Shaka uses TracksChangedEvent, while HLS.JS uses discrete audio and text track event AUDIO_TRACKS_UPDATED and SUBTITLE_TRACKS_UPDATED

src/player.js
OrenMe: I think because we decided to link ready to the player constructor it means this is something related to player creation, which is done only once, so don't think we can reuse it, so no point to check this, as you make sure to create it on constructor so you know it is always there.
If on changing a media application wants to listen to any player related event they can do so by registering to events, so does our UI.

src/event/events.js
dan-ziv: I think those are 2 separate events with different meanings.
TRACKS_PARSED called once when we have tracks for the first time, while TRACKS_CHANGED called in any change of tracks during playback.

src/player.js
yairans: I don't understand. the ready promise must be null on change media, else it will be resolved
before the new media will loaded. and on change media we don't create a new player and the constructor not invoked.

PR #64, opened 2017-06-19
feat(player): fire source selected event and expose session id setter

yairans:


src/player.js
```diff
@@ -144,6 +144,7 @@ class Player extends FakeEventTarget {
       let sources = config.sources;
       for (let i = 0; i < sources.length; i++) {
         if (Html5.canPlayType(sources[i].mimetype)) {
+          this.dispatchEvent(new FakeEvent(CustomEvents.SOURCE_SELECTED, {selectedSource: sources[i]}));
```
dan-ziv: @OrenMe @yairans 
We should send the whole source object and not only the url?

OrenMe: Good point, I agree

yairans: Kaltura player should decorate the url with the session id. therefore, the source has to passed by reference. primitive like string are passed by value.

PR #70, opened 2017-06-26
feat(playback config): handling of playback configuration and stream priority

dan-ziv:
- muted
- autoplay
- stream priority


src/player.js
```diff
@@ -130,51 +156,100 @@ class Player extends FakeEventTarget {
...
+   * Selects the engine to create based on a given configuration.
+   * @private
+   * @returns {boolean} - Whether a proper engine was found.
+   */
+  _selectEngine(): boolean {
+    let engineSelected = false;
+    if (this._config.sources) {
+      if (this._config.playback && this._config.playback.streamPriority) {
+        engineSelected = this._selectEngineByPriority();
+      } else {
```
yairans: we should have default priority

src/player.js
```diff
@@ -130,53 +158,95 @@ class Player extends FakeEventTarget {
    * @static
    */
   static _defaultConfig(): Object {
-    return {};
+    return DefaultPlayerConfig;
```
OrenMe: Make this immutable, we don't want default config to change.
you can just serialize it via object.assign

src/player.js
```diff
@@ -130,53 +158,95 @@ class Player extends FakeEventTarget {
...
   /**
-   * Select the engine to create based on the given configured sources.
-   * @param {Object} config - The configuration of the player instance.
+   * Selects the engine to create based on a given configuration.
+   * @private
+   * @returns {boolean} - Whether a proper engine was found.
+   */
+  _selectEngine(): boolean {
+    let engineSelected = false;
+    if (this._config.sources) {
```
OrenMe: We can always just do _selectEngineByPriority, cause we have the default setup.
Now just need to make sure the default config is deep cloned.

src/player.js
```diff
@@ -560,4 +644,5 @@ class Player extends FakeEventTarget {
 // </editor-fold>
 }
 
-export default Player;
+const registerEngine = Player.registerEngine;
+export {registerEngine};
```
OrenMe: Let's remove for now, don't see a use case for external playback engine that we want to support without too much added complexity.

src/player.js
dan-ziv: done

src/player.js
dan-ziv: done

src/player.js
dan-ziv: done

PR #79, opened 2017-07-09
fix(native-player): handle progressive sources

yairans:


src/utils/resolution.js
```diff
@@ -0,0 +1,50 @@
...
+        if (ratioDiff < minRatioDiff) {
+          minRatioDiff = ratioDiff;
+          mostSuitableWidthAndRatioTracks = [track];
+        } else if (ratioDiff === minRatioDiff) {
+          mostSuitableWidthAndRatioTracks.push(track);
+        }
+      }
+    }
+    let maxBandwidth = 0;
+    for (let track of mostSuitableWidthAndRatioTracks) { // select the top bitrate from the ratio filter results
```
itaykinnrot: @yairans  we should consider preferred bitrate as an external input no? @OrenMe

OrenMe: Yes, but all ABR section is not implemented yet, will be done as part of another PR that will handle it over all engines and adapters(native/hls/dash)

PR #81, opened 2017-07-18
feat(plugins): update plugins configuration on-the-fly

dan-ziv:


src/player.js
```diff
@@ -150,6 +151,24 @@ export default class Player extends FakeEventTarget {
   }
 
   /**
+   * Updates the configuration of the plugins.
+   * @param {Object} config - The player configuration.
+   * @private
+   * @returns {void}
+   */
+  _maybeConfigurePlugins(config: Object): void {
+    if (!Utils.Object.isEmptyObject(this._engine) && config.plugins) {
```
OrenMe: why do we need this check `!Utils.Object.isEmptyObject(this._engine)`? to support on the fly and avoid the initial double plugin setup?

PR #87, opened 2017-07-30
fix(player): fire source selected as array

yairans:


src/player.js
```diff
@@ -546,11 +547,11 @@ export default class Player extends FakeEventTarget {
 
   /**
    * Get the player config.
-   * @returns {Object} - The player configuration.
+   * @returns {Object} - A copy of the player configuration.
    * @public
    */
   get config(): Object {
-    return this._config;
+    return Utils.Object.mergeDeep({}, this._config);
```
OrenMe: @yairans better have this as separate PR as it won't appear in release notes otherwise

PR #106, opened 2017-09-07
feat: allow setting native hls and dash playback

OrenMe:

By default we use HLS and Dash adapters for playback of HLS and Dash streams.
This feature will allow setting the required adapter to be native or not, so we can use native HLS playback on Safari for example.

```json
{
  "playback": {
    "preferNative": {
      "hls": false,
      "dash": false
    }
  }
}
```

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -81,10 +81,10 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
    * @returns {boolean} - Whether the native adapter can play a specific mime type
    * @static
    */
-  static canPlayType(mimeType: string, preferNative: boolean): boolean {
+  static canPlayType(mimeType: string, preferNative: boolean = true): boolean {
     let canPlayType = false;
     if (typeof mimeType === 'string') {
-      canPlayType = !!(NativeAdapter.TEST_VIDEO.canPlayType(mimeType.toLowerCase())) && preferNative;
+      canPlayType = !!(NativeAdapter.TEST_VIDEO.canPlayType(mimeType.toLowerCase())) && !!preferNative;
     }
     NativeAdapter._logger.debug('canPlayType result for mimeType:' + mimeType + ' is ' + canPlayType.toString());
     return canPlayType;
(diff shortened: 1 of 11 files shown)


src/player-config.json
```diff
@@ -7,14 +7,16 @@
     "preload": "none",
     "autoplay": false,
     "muted": false,
-    "options": {
+    "tweaks": {
```
dan-ziv: @OrenMe  if we're changing this key we should change also the reference in dash & hls adapters

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -70,11 +77,15 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
    * Checks if NativeAdapter can play a given mime type.
    * @function canPlayType
    * @param {string} mimeType - The mime type to check
+   * @param {boolean} preferNative - prefer native flag
```
yairans: you can annotate the default value true
`@param {boolean} [preferNative=true] - prefer native flag
`

src/engines/html5/media-source/media-source-provider.js
```diff
@@ -65,15 +65,16 @@ export default class MediaSourceProvider {
   /**
    * Checks if the a media source adapter can play a given source.
    * @param {Source} source - The source object to check.
+   *  @param {boolean} preferNative - prefer native flag
```
yairans: as above

src/engines/html5/media-source/media-source-provider.js
```diff
@@ -95,7 +96,7 @@ export default class MediaSourceProvider {
   static getMediaSourceAdapter(videoElement: HTMLVideoElement, source: Source, config: Object): ?IMediaSourceAdapter {
     if (videoElement && source && config) {
       if (!MediaSourceProvider._selectedAdapter) {
-        MediaSourceProvider.canPlaySource(source);
+        MediaSourceProvider.canPlaySource(source, false);
```
yairans: why false?

src/player-config.json
```diff
@@ -8,14 +8,16 @@
...
```
yairans: i think 'options' is more intuitive...

src/engines/html5/media-source/adapters/native-adapter.js
OrenMe: Nice, didn't know this 👍

src/engines/html5/media-source/media-source-provider.js
OrenMe: why do we even have this check here? we should be getting here after we already have a selectedAdapter anyway.
I'll change to true for consistency, but this seems unnecessary.

yairans: This is api. You can asssume nothing how getting here. 
Why true/false. You have the param from config

PR #108, opened 2017-09-11
fix(FEC-7138): enable setting 0 and 1 values

OrenMe:

The volume check is not enabling setting 0 or 1

### CheckLists

- [ ] Docs have been updated


test/src/player.spec.js
```diff
@@ -1649,3 +1649,58 @@ describe('seekToLiveEdge', function () {
...
+  });
+
+  after(() => {
+    removeVideoElementsFromTestPage();
+    removeElement(targetId);
+  });
+
+  it('should return 1 by default', function () {
+    config.sources = sourcesConfig.Mp4;
+    player = new Player(targetId, config);
```
dan-ziv: lines 1674 and 1675 are repeated in each unit test,
so it's better to place them inside the beforeEach

PR #110, opened 2017-09-13
fix(FEC-7089): handle bitrate change in android browser

yairans:

On bitrate change for progressive download In Android browser (AKA native/stock browser) we have to seek to the current position only after some playback

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -383,7 +383,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
       this._eventManager.listen(this._videoElement, Html5Events.LOADED_DATA, () => {
         this._eventManager.unlisten(this._videoElement, Html5Events.LOADED_DATA);
         if (Env.browser.name === 'Android Browser') {
-          // Android browser doesn't support seek before play, we have to play wait to 'durationchanged' event and then seek.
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -381,14 +382,32 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
       this._sourceObj = videoTracks[videoTrack.index];
       this._eventManager.listen(this._videoElement, Html5Events.LOADED_DATA, () => {
         this._eventManager.unlisten(this._videoElement, Html5Events.LOADED_DATA);
-        this._eventManager.listen(this._videoElement, Html5Events.SEEKED, () => {
-          this._eventManager.unlisten(this._videoElement, Html5Events.SEEKED);
-          this._onTrackChanged(videoTrack);
-        });
-        this._videoElement.currentTime = currentTime;
+        if (Env.browser.name === 'Android Browser') {
```
OrenMe: @yairans can we make it use the actual data fro video and not browser sniffing based?
Is it possible to check after LOADED_DATA if video.duration exist and only if not then register DURATION_CHANGE and call play?
Also, if I recall this is a known bug that on Android stock browser the loadeddata event is fired before even the loadedmetadata is fired, can we move this implemntation to use loadedmetadata instead? if so then we need to test it again on all environments/platforms.

PR #113, opened 2017-09-14
feat: player doesn't attach itself to parent

OrenMe:

The application which uses the playkit-js SDK will need to use getView method to get the player element and it will control attaching it to the DOM

### CheckLists

- [x] Docs have been updated


OrenMe:
One test is failing @yairans please take a look as it might be something you are familiar with

yairans:
@OrenMe the failure test has been fixed

OrenMe:
@yairans do you know why we had the exception in promise? maybe this is a possible configuration path we need to protect?

PR #118, opened 2017-09-19
feat: caption refactor + support cvaa & default audio/text tracks

OrenMe:

* player handles synthetic VTT cue and displays them with a derived HTML Dom element that is created from this cue
* currently player support cues from native video element which are created via the adapters. Adapters themselves parse the VTT from whatever source(VTT segmented file, VTT in header, DFXP etc) and use the native textTrack API to add them to video and sync them to timeline. 
HTML5 layer gets the cuechange event and creates a synthetic VTT cue to normalize the VTTCue and TextTrackCue models to align with VTTCue one(thanks Edge and IE11) so this means we don't support VTT attributes fully in those browsers yet
* expose API to set text style and text cue display settings

### CheckLists

- [ ] Docs have been updated


OrenMe:
Still needs to:
* handle default text tracks from manifest
* external VTT files
* support VTT attributes from VTT file on edge and IE11
* use default texttrack API on iOS for native playback
* tests
* sync with UI elements, e.g. the control bar and possibly ads - need to change the text position when they appear

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+  }
+
+  /**
+   * update the text cue display settings
+   * @param {Object} cues - cues
+   * @param {Object} settings - text cue display settings
+   * @private
+   * @returns {void}
+   */
+  _updateCueDisplaySettings(cues: Array<any>, settings: Object){
```
dan-ziv: We are using `_updateCueDisplaySettings` twice and sending the method `this._lastUpdatedTextCues, this._textDisplaySettings`
Why not sending nothing and use the _this_ vars inside?

src/player.js
```diff
@@ -707,6 +846,48 @@ export default class Player extends FakeEventTarget {
...
+
+    if (playbackConfig.textLanguage == "auto"){
+      playbackConfig.textLanguage = Locale.language;
+    }
+
+    this._setDefaultTrack(TrackTypes.TEXT, playbackConfig.textLanguage, activeTracks.text);
+    this._setDefaultTrack(TrackTypes.AUDIO, playbackConfig.audioLanguage, activeTracks.audio);
+  }
+
+  _setDefaultTrack(type: string, language: string, defaultTrack: Track){
```
dan-ziv: I dont think that we need to send the type of the track.
We can check inside `_setDefaultTrack ` the track's instanceof.

src/player.js
```diff
@@ -707,6 +846,48 @@ export default class Player extends FakeEventTarget {
...
+      }));
+    }
+  }
+
+  _setDefaultTracks(){
+    const activeTracks = this.getActiveTracks();
+    const playbackConfig = this._config.playback;
+
+    if (playbackConfig.textLanguage == "auto"){
+      playbackConfig.textLanguage = Locale.language;
```
dan-ziv: Why do we change the user config? it will looks like a bug.
we should only do our inner logic according to the config which in this case is to send
`Locale.language` to `_setDefaultTrack ` .

src/player.js
```diff
@@ -707,6 +846,48 @@ export default class Player extends FakeEventTarget {
...
+  }
+
+  _setDefaultTrack(type: string, language: string, defaultTrack: Track){
+    if (language){
+      const track: ?Track = this._getTracksByType(type).find(track => track.language === language);
+      if (track) {
+        this.selectTrack(track);
+      }
+    } else if (defaultTrack) {
+      this.selectTrack(defaultTrack);
```
dan-ziv: This is already the selected track - why do we need to select it again?

src/engines/html5/html5.js
```diff
@@ -69,6 +85,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   constructor(source: Source, config: Object) {
     super();
+    this._config = config;
     this._eventManager = new EventManager();
     this._createVideoElement();
     this._loadMediaSourceAdapter(source, config);
```
dan-ziv: If we already save the config on the _this_ we don't need to send it to `_loadMediaSourceAdapter `

dan-ziv:
@OrenMe 
In general I think we should create TextTrackController and manage all the logic inside it

src/player.js
OrenMe: Yes, we don't want to select it necessarily cause we might have config from user that overrides the manifest default.
we only parse it from the adapter and set it as default, but don't actually select it and actual selection is done here where we resolve defaults

OrenMe:
@dan-ziv changing this to a controller is indeed the right way but it will require some changes that we will cover in other PR, that will handle adding some life cycle methods we currently lack.

src/engines/html5/html5.js
```diff
@@ -212,6 +302,10 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   hideTextTrack(): void {
     if (this._mediaSourceAdapter) {
       this._mediaSourceAdapter.hideTextTrack();
+      let textTrackEl: TextTrack = this._getSelectedTextTrack();
+      if (textTrackEl) {
+        textTrackEl.oncuechange = null;
+      }
```
yairans: just call `this. _removeCueChangeListener()`

src/engines/html5/html5.js
```diff
@@ -34,6 +35,21 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   _mediaSourceAdapter: ?IMediaSourceAdapter;
 
   /**
+   * Flag to indicate first time text track cue change.
+   * @member {Array<boolean>} - _first time
+   * @type {Array<boolean>}
+   * @private
+   */
+  _showTextTrackfirstTime: { [number]: boolean } = {};
```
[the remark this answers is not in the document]
yairans: also, you have to clean this map in destroy

src/engines/html5/html5.js
```diff
@@ -194,13 +212,85 @@ export default class Html5 extends FakeEventTarget implements IEngine {
 
   /**
    * Select a new text track.
-   * @param {TextTrack} textTrack - The text track object to set.
+   * @param {PKTextTrack} textTrack - The text track object to set.
```
yairans: The **playkit** text track object to set.

src/engines/html5/html5.js
```diff
@@ -194,13 +212,85 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+    this._removeCueChangeListener();
     if (this._mediaSourceAdapter) {
       this._mediaSourceAdapter.selectTextTrack(textTrack);
     }
+    this._addCueChangeListener(textTrack);
+  }
+
+  /**
+   * Add cuechange listener to active textTrack
+   * @param {PKTextTrack} textTrack - player text track
```
yairans: **playkit** text track

src/engines/html5/html5.js
```diff
@@ -194,13 +212,85 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+        textTrackEl.mode = this._showTextTrackfirstTime[textTrack.index] ? "hidden" : "showing";
+        this._showTextTrackfirstTime[textTrack.index] = true;
+        textTrackEl.oncuechange = (e) => this._onCueChange(e);
+      }
+    }
+  }
+
+  /**
+   * Remove cuechange listener to active textTrack
+   * @param {PKTextTrack} textTrack - player text track
```
yairans: **playkit** text track

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -440,15 +440,16 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
   /**
    * Select a text track
    * @function selectTextTrack
-   * @param {TextTrack} textTrack - the track to select
+   * @param {PKTextTrack} textTrack - the track to select
```
yairans: the **playkit text** track to select

yairans:
remove dist

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
   }
 
   /**
+   * Sets style attributes for text tracks.
+   * @param {Object} style - text styling settings
```
yairans: `{TextStyle}`

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+      Utils.Dom.appendChild(document.head, element);
+    }
+    let sheet = element.sheet;
+
+    while (sheet.cssRules.length) {
+      sheet.deleteRule(0);
+    }
+
+    if (this._config.playback.useNativeTextTrack) {
+      sheet.insertRule(`video.${ENGINE_CLASS_NAME}::cue { ${style.toCSS()} }`, 0);
```
yairans: is it valid in ie11?

yairans: check if `this._config.playback` and `style` is exists and has `toCSS()` method

src/player.js
```diff
@@ -130,6 +146,30 @@ export default class Player extends FakeEventTarget {
...
+   * @type {string}
+   * @private
+   */
+  _playerId: string;
+  /**
+   * The player last updated text cues list
+   * @type {any}
+   * @private
+   */
+  _activeTextCues: Array<any> = [];
```
yairans: should clean on destroy

src/player.js
```diff
@@ -130,6 +146,30 @@ export default class Player extends FakeEventTarget {
...
+  /**
+   * The player text disaply settings
+   * @type {Object}
+   * @private
+   */
+  _textDisplaySettings: Object = {};
```
yairans: should clean on destroy

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+    }
+  }
+
+  /**
+   * update the text display settings
+   * @param {Object} settings - text cue display settings
+   * @public
+   * @returns {void}
+   */
+  setTextDisplaySettings(settings: Object): void {
```
yairans: is this for height and `setTextStyle ` for color etc.? not clear enough.

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+
+  /**
+   * update the text cue display settings
+   * @private
+   * @returns {void}
+   */
+  _updateCueDisplaySettings(){
+    const activeCues = this._activeTextCues;
+    const settings= this._textDisplaySettings;
+    for (let i = 0; i< activeCues.length; i++ ){
```
yairans: check `activeCues` and `settings` existence.

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+  /**
+   * update the text display settings
+   * @param {Object} settings - text cue display settings
+   * @public
+   * @returns {void}
+   */
+  setTextDisplaySettings(settings: Object): void {
+    this._textDisplaySettings = settings;
+    this._updateCueDisplaySettings();
+    for (let i = 0; i < this._activeTextCues.length; i++) {
```
yairans: better to check `_activeTextCues` existence. i know you init as [] in  html5 `_onCueChange` but still....

src/player.js
```diff
@@ -130,6 +146,30 @@ export default class Player extends FakeEventTarget {
    */
   _el: HTMLDivElement;
   /**
+   * The player text DOM element container.
+   * @type {HTMLDivElement}
+   * @private
+   */
+  _textDisplayEl: HTMLDivElement;
```
yairans: should clean on destroy ? or can rely vid element destroy?

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+    this._updateCueDisplaySettings();
+    for (let i = 0; i < this._activeTextCues.length; i++) {
+      this._activeTextCues[i].hasBeenReset = true;
+    }
+    this._updateTextDisplay(this._activeTextCues);
+  }
+
+  /**
+   * handle text cue change
+   * @param {Event} event - the cue change event payload
```
yairans: {FakeEvent}

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+    this._updateCueDisplaySettings();
+    this._updateTextDisplay(this._activeTextCues)
+  }
+
+  /**
+   * update the text cue display settings
+   * @private
+   * @returns {void}
+   */
+  _updateCueDisplaySettings(){
```
yairans: : void

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+    }
+  }
+
+  /**
+   * update the text display
+   * @param {Object} cues - list of cues
+   * @private
+   * @returns {void}
+   */
+  _updateTextDisplay(cues: any): void {
```
yairans: should be `Array<Cue>`. also in jsdoc

src/engines/html5/html5.js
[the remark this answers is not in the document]
OrenMe: No need to clean it, after a text track is shown for first time it is reused, and no need for this hack

src/engines/html5/html5.js
OrenMe: ooopps - I don't even use it so no ned to define it

src/player.js
```diff
@@ -227,6 +267,90 @@ export default class Player extends FakeEventTarget {
...
+    if (this._textDisplayEl === undefined) {
+      this._textDisplayEl = Utils.Dom.createElement("div");
+      Utils.Dom.addClassName(this._textDisplayEl, SUBTITLES_CLASS_NAME);
+      Utils.Dom.appendChild(this._el, this._textDisplayEl);
+    }
+    processCues(window, cues, this._textDisplayEl);
```
yairans: maybe add try/catch around `processCues`

src/engines/html5/html5.js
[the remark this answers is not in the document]
yairans: change media?

src/player.js
OrenMe: they are initialized and typed and only internal, so no reason for extra validation

src/engines/html5/html5.js
```diff
@@ -194,13 +212,85 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+  /**
+   * Get currently selected text track
+   * @returns {*} - returns the active text track element if available
+   * @private
+   */
+  _getSelectedTextTrack(): ?TextTrack {
+    const textTracks = this._el.textTracks;
+    for (let track in textTracks) {
+      if (textTracks.hasOwnProperty(track)) {
+        if (textTracks[parseInt(track)] && textTracks[parseInt(track)].mode !== "disabled") {
```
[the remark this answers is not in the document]
OrenMe: Regarding the mode, I added in the `_setDefaultTracks` a call to `hideTextTrack` so it will take care of setting them to disabled.

src/player.js
OrenMe: No need, we clean the content on reset and this is just the placeholder

src/player.js
OrenMe: it's not valid on ie11, but we use it only on iOS and the actual setup is done in application level that should check when it wants to use it and then just set `useNativeTextTrack: true`, otherwise use the DIV with CSS styling

OrenMe:
@dan-ziv @yairans please review following review fixes

src/player.js
OrenMe: why? it's a managed method, the processCues has it's own protections internally

src/engines/html5/html5.js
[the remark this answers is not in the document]
yairans: `hideTextTrack` calls the adapter `hideTextTrack`. u have to make sure the adapter hides by `disabled` and not by `hidden`. currently is by `hidden`
https://github.com/kaltura/playkit-js-hls/blob/c7a9ef53046d004fc7b4f021bfd275132484ee19/src/hls-adapter.js#L443

src/engines/html5/html5.js
dan-ziv: done

src/player.js

src/engines/html5/html5.js
```diff
@@ -282,7 +288,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @returns {?TextTrack} - returns the active text track element if available
    * @private
    */
-  _getSelectedTextTrack(): ?TextTrack {
+  _getSelectedTextTrackElement(): ?TextTrack {
```
OrenMe: It's not an element, but a DOM object.
if it was of the format then I would agree.
```html
<video>
   <track>...</track>
</video>
```

And even for the above, you can still get the element as an object via `video.textTracks` and via DOM API selector.

OrenMe:
added tests

PR #122, opened 2017-09-25
feat: configure refactor

dan-ziv:

Align configure api behavior:

**Plugins:**
1. If engine is already created, the existing plugins can update their configuration on the fly. If a new plugin is configured after engine is exists, we will ignore this request.
2. If engine is not yet created, plugins can be either loaded or update their configuration at any time.

**Playback:**
Playback can be configured at any time but this will affect only to until **sources will be configured**.
For example, If sources not yet configured and playback is configured, the playback configuration will take place when sources will received.
If sources already received and playback is configured again, the playback configuration will take place on the next sources configuration. 
### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -363,7 +363,15 @@ export default class Player extends FakeEventTarget {
         this.playsinline = true;
       }
       if (this._config.playback.preload === "auto") {
-        this.load();
+        /**
+         * If ads plugin enabled it's his responsibility to preload the content player.
+         * So to avoid loading the player twice which can cause errors on MSEs we are not
+         * calling load from the player.
+         * TODO: Change it to check the ads configuration when we will develop the ads manager.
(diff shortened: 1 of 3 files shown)


src/player.js
```diff
@@ -182,25 +181,72 @@ export default class Player extends FakeEventTarget {
...
+
+  /**
+   * Updates the player configuration.
+   * @param {Object} config - The new received configuration.
+   * @private
+   * @returns {void}
+   */
+  _updatePlayerConfig(config: Object): void {
+    if (Utils.Object.isEmptyObject(this._config)) {
+      Utils.Object.mergeDeep(this._config, Player._defaultConfig, config);
```
OrenMe: Why no do this on init anyway?

PR #123, opened 2017-09-25
fix(base-plugin): getConfig api should return copy of his config

dan-ziv:

getConfig api should return copy of his config instead of reference.

### CheckLists

- [ ] Docs have been updated


OrenMe:
@dan-ziv better add tests here before merge

dan-ziv:
@OrenMe 
Test added.

PR #124, opened 2017-09-28
feat: change media

dan-ziv:

* Change media added logic.
* Requires to add reset() api to most of our components.
* Requires to change destroy API of engine and MSE adapters to return promises (since shaka destroy API returns promise).
* Lifecycle events of change media  were added.

### CheckLists

- [x] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -46,7 +46,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   _showTextTrackFirstTime: { [number]: boolean } = {};
   /**
-   * Flag to indicate when a media source adapter can be loaded.
+   * Promise to indicate when a media source adapter can be loaded.
    * @type {Promise<*>}
    * @private
    */
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -117,7 +117,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
         this.detach();
         if (this._el) {
           this.pause();
-          this._el.removeAttribute('src');
+          Utils.Dom.removeAttribute(this._el, 'src');
           Utils.Dom.removeChild(this._el.parentNode, this._el);
         }
         this._eventManager.destroy();
@@ -141,7 +141,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
         this.detach();
         this._eventManager.removeAll();
         if (this._el) {
-          this._el.removeAttribute('src');
+          Utils.Dom.removeAttribute(this._el, 'src');
         }
         return (this._mediaSourceAdapter ? this._mediaSourceAdapter.destroy() : Promise.resolve())
           .then(() => {
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -91,6 +91,13 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     this._init(source, config)
   }
 
+  /**
+   * Initializes the engine in case no media source adapter is attached.
+   * @param {Source} source - The selected source object.
+   * @param {Object} config - The player configuration.
+   * @private
+   * @returns {void}
+   */
   _init(source: Source, config: Object): void {
     if (!this._mediaSourceAdapter) {
       this._config = config;
@@ -102,31 +109,42 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   /**
    * Destroys the engine.
    * @public
-   * @returns {void}
+   * @returns {Promise<*>} - The destroy promise.
    */
-  destroy(): void {
-    this.detach();
-    if (this._mediaSourceAdapter) {
-      this._mediaSourceAdapter.destroy();
-      this._mediaSourceAdapter = null;
+  destroy(): Promise<*> {
+    return new Promise((resolve) => {
+      this.detach();
+      if (this._el) {
+        this.pause();
+        this._el.removeAttribute('src');
+        Utils.Dom.removeChild(this._el.parentNode, this._el);
+      }
+      this._eventManager.destroy();
       MediaSourceProvider.destroy();
-    }
-    if (this._el) {
-      this.pause();
-      this._el.removeAttribute('src');
-      Utils.Dom.removeChild(this._el.parentNode, this._el);
-    }
-    this._eventManager.destroy();
-    instance = null;
+      (this._mediaSourceAdapter ? this._mediaSourceAdapter.destroy() : Promise.resolve())
+        .then(() => {
+          this._mediaSourceAdapter = null;
+          instance = null;
+          resolve();
+        });
+    });
   }
 
-  reset(): void {
-    this.detach();
-    if (this._mediaSourceAdapter) {
-      this._mediaSourceAdapter.destroy();
-      this._mediaSourceAdapter = null;
-    }
-    this._eventManager.removeAll();
+  /**
+   * Resets the engine.
+   * @public
+   * @returns {Promise<*>} - The reset promise.
+   */
+  reset(): Promise<*> {
+    return new Promise((resolve) => {
+      this.detach();
+      this._eventManager.removeAll();
+      (this._mediaSourceAdapter ? this._mediaSourceAdapter.destroy() : Promise.resolve())
+        .then(() => {
+          this._mediaSourceAdapter = null;
+          resolve();
+        });
+    });
   }
 
   /**
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -29,7 +29,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   _eventManager: EventManager;
   /**
    * The selected media source adapter of the engine.
-   * @type {IMediaSourceAdapter}
+   * @type {?IMediaSourceAdapter}
    * @private
    */
   _mediaSourceAdapter: ?IMediaSourceAdapter;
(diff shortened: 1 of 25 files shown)


dan-ziv:
TODOS:

- [ ] Check DRM entries.
- [ ]  Check with ads.

src/engines/html5/html5.js
```diff
@@ -111,11 +162,11 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @returns {void}
    */
   attach(): void {
-    for (let playerEvent in Html5Events) {
-      this._eventManager.listen(this._el, Html5Events[playerEvent], () => {
-        this.dispatchEvent(new FakeEvent(Html5Events[playerEvent]));
+    Object.keys(Html5Events).forEach((html5Event) => {
```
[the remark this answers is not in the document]
dan-ziv: I hate `for in` since eslint shout on you if u don't include the `hasOwnProperty` protection.
This is seems like two actions added, but `for in` do exactly the same behind the scene.

BTW, u can run the tests here:
https://jsperf.com/object-keys-vs-for-in-with-closure/3
and see the `Object.keys().forEach()` is **much more** faster.

src/engines/html5/html5.js
```diff
@@ -130,10 +181,10 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @returns {void}
    */
   detach(): void {
-    for (let playerEvent in Html5Events) {
-      this._eventManager.unlisten(this._el, Html5Events[playerEvent]);
-    }
-    if (this._mediaSourceAdapter) { // unlisten to adaptive bitrate changed
+    Object.keys(Html5Events).forEach((html5Event) => {
```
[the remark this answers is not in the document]
dan-ziv: same as above comment

src/player.js
```diff
@@ -202,53 +220,49 @@ export default class Player extends FakeEventTarget {
...
-        // We allow to load plugins as long as the player has no engine
-        if (!this._engine) {
-          this._pluginManager.load(name, this, plugins[name]);
-          let plugin = this._pluginManager.get(name);
-          if (plugin) {
-            this._config.plugins[name] = plugin.getConfig();
-            if (typeof plugin.getMiddlewareImpl === "function") {
-              this._playbackMiddleware.use(plugin.getMiddlewareImpl());
+    if (plugins) {
+      Object.keys(plugins).forEach((name) => {
```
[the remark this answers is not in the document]
dan-ziv: Yes, it can be empty.
Pay attention that the received arg of this method is the **new** plugins that received from the configure and for that it can be null.

src/player.js
```diff
@@ -184,14 +191,25 @@ export default class Player extends FakeEventTarget {
...
     if (config.sources) {
-      this._maybeResetPlayer();
-      if (this._selectEngineByPriority()) {
-        this._appendEngineEl();
-        this._posterManager.setSrc(this._config.metadata.poster);
-        this._posterManager.show();
-        this._attachMedia();
-        this._handlePlaybackConfig();
+      const receivedSourcesWhenHasEngine: boolean = !!this._engine;
+      if (receivedSourcesWhenHasEngine) {
```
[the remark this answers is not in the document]
dan-ziv: I don't see any reason to fire SOURCE_CHANGED when u don't have yet any sources.

src/player.js
```diff
@@ -271,16 +285,20 @@ export default class Player extends FakeEventTarget {
    * @public
    */
   destroy(): void {
-    if (this._engine) {
-      this._engine.destroy();
-    }
+    this._canLoadEnginePromise = this._engine ? this._engine.reset() : Promise.resolve();
```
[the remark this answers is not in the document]
dan-ziv: Why? do u see any use case for this?
If we already decide to do something like this, lets expose destroy() and reset() of the player as promises (this promise exists to avoid exactly this).

src/player.js
```diff
@@ -343,11 +364,11 @@ export default class Player extends FakeEventTarget {
    */
   _attachMedia(): void {
     if (this._engine) {
-      for (let playerEvent in Html5Events) {
-        this._eventManager.listen(this._engine, Html5Events[playerEvent], (event: FakeEvent) => {
+      Object.keys(Html5Events).forEach((html5Event) => {
```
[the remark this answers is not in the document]
dan-ziv: Believe me, u don't want to... :)
https://jsperf.com/object-keys-vs-for-in-with-closure/3

src/engines/html5/html5.js
```diff
@@ -32,22 +38,33 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+   * Factory method to get an engine.
    * @param {Source} source - The selected source object.
    * @param {Object} config - The player configuration.
    * @returns {IEngine} - New instance of the run time engine.
    * @public
    * @static
    */
-  static createEngine(source: Source, config: Object): IEngine {
-    return new this(source, config);
+  static getEngine(source: Source, config: Object): IEngine {
```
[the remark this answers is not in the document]
dan-ziv: This is internal managed by us, not sure I understand you

src/player.js
```diff
@@ -184,14 +191,25 @@ export default class Player extends FakeEventTarget {
     Utils.Object.mergeDeep(this._config, config);
     this._configureOrLoadPlugins(config.plugins);
     if (config.sources) {
-      this._maybeResetPlayer();
```
yairans: `config.source` could be empty. `[]` consider as `true`

src/player.js
```diff
@@ -327,13 +343,16 @@ export default class Player extends FakeEventTarget {
...
    * Loads the selected engine.
-   * @param {IEngine} engine - The selected engine.
+   * @param {IEngine} Engine - The selected engine.
    * @param {Source} source - The selected source object.
    * @private
    * @returns {void}
    */
-  _loadEngine(engine: typeof IEngine, source: Source): void {
-    this._engine = engine.createEngine(source, this._config);
+  _loadEngine(Engine: typeof IEngine, source: Source) {
```
yairans: `: void`

src/plugin/base-plugin.js
```diff
@@ -123,6 +123,17 @@ export default class BasePlugin implements IPlugin {
...
 
   /**
+   * Runs the reset logic of the plugin.
+   * plugin must implement this method.
+   * @public
+   * @abstract
+   * @returns {void}
+   */
+  reset(): void {
+    throw new PlayerError(PlayerError.TYPE.NOT_IMPLEMENTED_METHOD, 'reset()').getError();
```
yairans: maybe a plugin has nothing to do in `reset`. better to leave it empty.

src/player.js
dan-ziv: sources is an object

src/engines/html5/html5.js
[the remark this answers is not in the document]
OrenMe: get means get, not init.
If you must call reset to get the init working correctly then we have an unmanaged code path

src/player.js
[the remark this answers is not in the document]
OrenMe: changing source from null to a source is still a change, and it means we act consistently.
let's always emit SOURCE_CHANGED when sources are sent to configure

src/player.js
yairans: same potential problem in object

src/engines/html5/html5.js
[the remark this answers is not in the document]
dan-ziv: Do u have any suggestion?

dan-ziv:
@OrenMe @yairans 
The PR has been updated with new design.
Please review again.
Thanks

src/plugin/base-plugin.js
dan-ziv: I think it the same as 'destroy' - we need to force plugins to implement those methods

src/engines/html5/html5.js
```diff
@@ -84,11 +88,27 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   constructor(source: Source, config: Object) {
     super();
-    this._config = config;
     this._eventManager = new EventManager();
+    this._canLoadMediaSourceAdapterPromise = Promise.resolve();
     this._createVideoElement();
-    this._loadMediaSourceAdapter(source);
-    this.attach();
+    this._init(source, config)
```
yairans: missing `;`

src/engines/html5/html5.js
```diff
@@ -84,11 +88,27 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+   * @returns {void}
+   */
+  restore(source: Source, config: Object): void {
+    this.detach();
+    this._eventManager.removeAll();
+    if (this._el) {
+      Utils.Dom.removeAttribute(this._el, 'src');
+    }
+    this._canLoadMediaSourceAdapterPromise = (this._mediaSourceAdapter ? this._mediaSourceAdapter.destroy() : Promise.resolve());
+    this._mediaSourceAdapter = null;
```
yairans: move these 2 lines to `init` and remove also from constructor

src/engines/html5/html5.js
```diff
@@ -33,20 +33,24 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @private
```
yairans: should be `@type {?IMediaSourceAdapter}`

src/engines/html5/html5.js
```diff
@@ -84,11 +88,27 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   constructor(source: Source, config: Object) {
     super();
-    this._config = config;
     this._eventManager = new EventManager();
+    this._canLoadMediaSourceAdapterPromise = Promise.resolve();
```
yairans: remove. taken from init. see comment below

src/engines/html5/html5.js
dan-ziv: fixed

PR #127, opened 2017-10-04
patch: initialization of textStyle object

dvh91:

textStyle object initialized so the UI could understand what the initial textStyle is and show accordingly the captions style options.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -176,7 +176,7 @@ export default class Player extends FakeEventTarget {
    * @type {TextStyle}
    * @private
    */
-  _textStyle: TextStyle = new TextStyle();
+  _textStyle: TextStyle;
   /**
    * The playback middleware of the player.
    * @type {PlaybackMiddleware}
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -176,7 +176,7 @@ export default class Player extends FakeEventTarget {
    * @type {TextStyle}
    * @private
    */
-  _textStyle: TextStyle;
+  _textStyle: TextStyle = new TextStyle();
```
dan-ziv: Move initialization to constructor

OrenMe:
@dvh91 don't know if you deleted the default Pull Request template or it didn't appear but we have it there for a reason so please add it to the first comment of this PR(just edit it) and add details and tick the corresponding check boxes after you save it.
Thanks!

OrenMe:
@dan-ziv ES6 classes member initialization can be done like this in general, and it is exactly the same as constructor initialization AFAIK.
Only issue is the fact we do some in constructor and some in member definition, so there's no alignment.
By the way, if you check the actual transpiled class you will see it does the _textTrack init in the class function definition, e.g.
```javascript
function Player() {
   ...
   this._textTrack = new ...
   ...
}
```

dvh91:
@OrenMe - Updated the first comment with the pull request template.

dan-ziv:
@OrenMe 
I know that,
The reason I asked this is only for consistency.

PR #128, opened 2017-10-08
fix(FEC-7226, FEC-7243): create comparer func to default tracks

dan-ziv:

The comparer func will consider the length of the inputs and return the appropriate comparison accordingly.  

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1228,8 +1228,16 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _setDefaultTrack(type: string, language: string, defaultTrack: Track): void {
+    function langComparer(inputLang: string, trackLang: string): boolean {
+      if (inputLang.length === trackLang.length) {
+        return inputLang === trackLang;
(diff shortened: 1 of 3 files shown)


src/player.js
```diff
@@ -1228,8 +1228,24 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _setDefaultTrack(type: string, language: string, defaultTrack: Track): void {
+    /**
+     * Comparing language string according to their length.
+     * @param {string} inputLang - The configured language.
+     * @param {string} trackLang - The default track language.
+     * @returns {boolean} - Whether the strings are equal or starts with the same substring.
+     */
+    function langComparer(inputLang: string, trackLang: string): boolean {
```
OrenMe: if this is a utility for text track so move it to textTrack class as static mehod and use it:
```javascript
TextTrack.langComparer(language, track.language)
```

PR #129, opened 2017-10-08
fix(FEC-7071): call pause on ended for browsers which don't do it natively

yairans:

IE11 and some old safari's don't pause on ended. so do it for them.

### CheckLists

- [x] Docs have been updated


OrenMe:
that's a bummer, browser variance is s$@t.
On a side note to anyone doing something related - need to make sure we don't depend on ended and pause event sequence in other places as it will be the other way around in those cases covered by this fix.

PR #130, opened 2017-10-08
feat: add default text font family and expose ENUM

OrenMe:

add default text font-family to text style and expose the TextStyle on the player instance so ENUM is more accessible.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/track/text-style.js
+++ b/src/track/text-style.js
@@ -21,11 +21,8 @@ class TextStyle {
    */
   static FontFamily: {[string]: string} = {
     "ARIAL": "Arial",
-    "ROBOTO": "Roboto",
-    "ARIAL_UNICODE_MS": "Arial Unicode Ms",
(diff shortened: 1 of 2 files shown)


src/track/text-style.js
```diff
@@ -15,12 +15,27 @@
 class TextStyle {
 
   /**
+   * Defined set of font families
+   * @enum {Object.<string, string>}}
+   * @export
+   */
+  static FontFamily: {[string]: string} = {
+    "ARIAL": "Arial",
+    "ROBOTO": "Roboto",
```
dvh91: I wouldn't expose Roboto and other custom non-system fonts in the enum.
Maybe Hillel can assist with the font choice but I would go with:
'sans-serif' and 'serif'
other than that, the custom skin can define his own custom fonts and add it to the font options.

PR #132, opened 2017-10-09
feat: order media source adapters depends on preferNative value

dan-ziv:

Change the implementation of preferNative logic for order the MSEs array, instead of return always false in NativeAdapter when preferNative=false.

### CheckLists

- [ ] Docs have been updated


OrenMe:
Great, way better then before.
We can even further optimize it if push comes to shove and enable this way a full adapters reordering based on other conditions or even actual name

PR #133, opened 2017-10-10
fix(FEC-7228): fixing default captions flow

odedhutzler:

Changed the captions flow so it will be:
if there is language defined in the configuration then select it.
if the language is "auto": and there is default language than take it else take the locale from the browser.
if there is no language defined ("") select "off"

### CheckLists

- [x] Docs have been updated


src/player.js
```diff
@@ -1224,11 +1241,25 @@ export default class Player extends FakeEventTarget {
...
+        textLanguage = activeTracks.text.language;
+      } else {
+        textLanguage = Locale.language;
+      }
+    } else if (configTextLang !== '') {
+        textLanguage = configTextLang;
+    }
+
+
+    //const textLanguage = (playbackConfig.textLanguage === "auto") ? Locale.language : playbackConfig.textLanguage;
```
dan-ziv: remove commented line

src/player.js
```diff
@@ -1224,11 +1241,25 @@ export default class Player extends FakeEventTarget {
   _setDefaultTracks(): void {
     const activeTracks = this.getActiveTracks();
     const playbackConfig = this._config.playback;
+    const configTextLang = playbackConfig.textLanguage;
+    const fallbackTextTrack: ?Track =  this._getTracksByType(TrackTypes.TEXT).find(track => TextTrack.langComparer(OFF, track.language));
+    let textLanguage = '';
```
dan-ziv: replace to const

src/player.js
```diff
@@ -1224,11 +1241,25 @@ export default class Player extends FakeEventTarget {
   _setDefaultTracks(): void {
     const activeTracks = this.getActiveTracks();
     const playbackConfig = this._config.playback;
```
dan-ziv: don't refer to the config via the private _config since you'll get reference and not copy and it might be dangerous.
`const playbackConfig = this.config.playback;`

src/player.js
```diff
@@ -1224,11 +1241,25 @@ export default class Player extends FakeEventTarget {
...
 
-    const textLanguage = (playbackConfig.textLanguage === "auto") ? Locale.language : playbackConfig.textLanguage;
-    this._setDefaultTrack(TrackTypes.TEXT, textLanguage, activeTracks.text);
+    if (configTextLang === AUTO) {
+      if (activeTracks.text) {
+        textLanguage = activeTracks.text.language;
+      } else {
+        textLanguage = Locale.language;
+      }
+    } else if (configTextLang !== '') {
```
dan-ziv: and you'll not need this condition
if textLanguage is not auto, it will be the explicit lang if user defined this or empty (because of the player default config) if not.

src/player.js
dan-ziv: init it as the value of `playbackConfig.textLanguage;`

src/player.js
```diff
@@ -1224,11 +1241,25 @@ export default class Player extends FakeEventTarget {
...
     const playbackConfig = this._config.playback;
+    const configTextLang = playbackConfig.textLanguage;
+    const fallbackTextTrack: ?Track =  this._getTracksByType(TrackTypes.TEXT).find(track => TextTrack.langComparer(OFF, track.language));
+    let textLanguage = '';
 
     this.hideTextTrack();
 
-    const textLanguage = (playbackConfig.textLanguage === "auto") ? Locale.language : playbackConfig.textLanguage;
-    this._setDefaultTrack(TrackTypes.TEXT, textLanguage, activeTracks.text);
+    if (configTextLang === AUTO) {
```
dan-ziv: if you will init the textLanguage as above, now just manipulate and check textLanguage

dan-ziv:
@odedhutzler 
In general I propose to move all the text language selection logic to private method that returns the text language

src/player.js
```diff
@@ -1223,15 +1240,38 @@ export default class Player extends FakeEventTarget {
...
+
+  /**
+   * Gets the text value that should be set in the captions - from the browser / activeTrack if the confiuuration is Auto
+   * @param {Object} activeTracks - active track data
+   * @param {string} textLanguage - the language, in case the configuration is not auto
+   * @private
+   * @returns {string}
+   */
+
+  _getTextLanguage(activeTracks: Object, textLanguage: string): string {
```
OrenMe: better call this `_getLanguage` and pass it the `activeTracks.text`. 
This way we can use it generically for both text and audio.

Also remember to change the flow type check to be of type Track instead of Object.

OrenMe: Need to check what happens for default audio as well.
The `Auto` should apply for it as well.
Does audio always come back with `activeTracks.audio` after parsing the manifest? if so the Locale option won't be valid in any case - @odedhutzler need to check this with both HLS on Safari, HLS with HLS.JS and Dash with Shaka.

OrenMe:
@odedhutzler we are not spec compliant.
Auto option was meant to behave as follow - 
1. Take Locale
2. Take manifest default
3. Take first language in manifest

src/player.js
```diff
@@ -1239,15 +1256,46 @@ export default class Player extends FakeEventTarget {
...
+   * @private
+   * @returns {string}
+   */
+
+  _getLanguage(language: string, activeTrack: ?Track, type: string): string {
+
+
+    if (language === AUTO) {
+      const localLanguage = Locale.language;
+      const textTracks = this._getTracksByType(type);
```
dan-ziv: if this is a generic function u can't call it `textTracks`

src/player.js
```diff
@@ -1239,15 +1256,46 @@ export default class Player extends FakeEventTarget {
...
+   * @param {string} language - the language, in case the configuration is not auto
+   * @param {Track} activeTracks - active track data
+   * @param {string} type - the track type (text / audio)
+   * @private
+   * @returns {string}
+   */
+
+  _getLanguage(language: string, activeTrack: ?Track, type: string): string {
+
+
```
dan-ziv: remove redundant lines

src/player.js
```diff
@@ -1239,15 +1256,46 @@ export default class Player extends FakeEventTarget {
...
+    if (language === AUTO) {
+      const localLanguage = Locale.language;
+      const textTracks = this._getTracksByType(type);
+      const localeTrack: ?Track = textTracks.find(track => TextTrack.langComparer(localLanguage, track.language));
+
+      if (localeTrack) {
+        language = localeTrack.language;
+      } else if (activeTrack && Object.keys(activeTrack).length) {
```
dan-ziv: why the 
` Object.keys(activeTrack).length`
check?

src/player.js
```diff
@@ -1239,15 +1256,46 @@ export default class Player extends FakeEventTarget {
...
+        language = activeTrack.language;
+      } else {
+        language = textTracks[0].language;
```
dan-ziv: also need to check that `textTrack` exist and has at least one track
`else if (textTrack && trackTracks.length > 0)`

src/player.js
```diff
@@ -1239,15 +1256,46 @@ export default class Player extends FakeEventTarget {
...
+      }
+
```
dan-ziv: remove redundant line

src/player.js
OrenMe: @odedhutzler calling it textTrack beats the purpose of this being a generic method

odedhutzler: yes, Dan also mentioned it :)

src/player.js
dan-ziv: No, it can be undefined and therefore the first condition is enough.

Btw, to check empty object use Utils.Object.isEmptyObject and not like u did

PR #136, opened 2017-10-19
fix(FEC-7242): subtitles transition in safari 9.0

odedhutzler:

subtitles are dropping from the top of the player to their correct location in safari 9. it looks like animation.. the correct behaviour would be that the subtitles appears at the bottom of the player.

@OrenMe - seems like there is duplication in the css.. why is there a css file for the subtitles in the playkit-js? isn't all the ui handled in the playkit-ui-js? should i try and remove the whole file?

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -49,9 +49,3 @@
   left: 0;
 }
 
-.playkit-subtitles > div > div {
-  -webkit-transition: 0.3s ease-in-out;
-  -moz-transition: 0.3s ease-in-out;
-  -o-transition: 0.3s ease-in-out;
-  transition: 0.3s ease-in-out;
-}


src/assets/style.css
```diff
@@ -49,9 +49,3 @@
   left: 0;
 }
 
-.playkit-subtitles > div > div {
-  -webkit-transition: 0.3s ease-in-out;
-  -moz-transition: 0.3s ease-in-out;
```
dan-ziv: @odedhutzler 
From what remember Safari transition is only the `-moz-` transition (needs to verify that), so why remove all?

odedhutzler: 1. Safari is webkit, moz is for mozilla (Firefox). Removing everything because i don't understand why we handle this here - i think it should be handled in the playkit-js-ui. more than that - i think the content of this file should move to the playkit-js-ui.

2. of course.

dan-ziv: Great, this is why I wrote that need to verify, so need to remove the webkit.
The core is in charge of displaying the captions and provide api to change their styles, the ui is in charge to use this api.

odedhutzler: ah - so this should be checked without the ui, to see this change doesn't mess the default captions.

kaltura-hooks:
Hi @odedhutzler,

PR #138, opened 2017-10-19
feat(cvaa): add TEXT_STYLE_CHANGED event

yairans:
Fire an event on text style changed, mainly for local storage.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -794,7 +794,7 @@ export default class Player extends FakeEventTarget {
         sheet.insertRule(`#${this._playerId} .${SUBTITLES_CLASS_NAME} > div > div > div { ${style.toCSS()} }`, 0);
       }
       this._textStyle = style;
-      this.dispatchEvent(new FakeEvent(CustomEvents.TEXT_STYLE_CHANGED, {textStyle: style}));
+      this.dispatchEvent(new FakeEvent(CustomEvents.TEXT_STYLE_CHANGED, {textStyle: this._textStyle.clone()}));
     } catch (e) {
       Player._logger.error(e.message);
     }
@@ -1031,9 +1031,6 @@ export default class Player extends FakeEventTarget {
       }
       if (typeof this._config.playback.playsinline === 'boolean') {
         this.playsinline = this._config.playback.playsinline;
-      }
-      if (typeof this._config.playback.textStyle === "object") {
-        this.textStyle = Utils.Object.mergeDeep(new TextStyle(), this._config.playback.textStyle);
       }
       if (this._config.playback.preload === "auto") {
         /**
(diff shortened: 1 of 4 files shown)


src/player.js
```diff
@@ -794,6 +794,7 @@ export default class Player extends FakeEventTarget {
         sheet.insertRule(`#${this._playerId} .${SUBTITLES_CLASS_NAME} > div > div > div { ${style.toCSS()} }`, 0);
       }
       this._textStyle = style;
+      this.dispatchEvent(new FakeEvent(CustomEvents.TEXT_STYLE_CHANGED, {textStyle: style}));
```
OrenMe: Don’t emit object by ref, in this case don’t emit the event with payload at all, make the listener of this explicitly need to call player.textStyle to get it and remove this payload as this is not data but an actual class instance object

src/player.js
```diff
@@ -1031,6 +1032,9 @@ export default class Player extends FakeEventTarget {
       if (typeof this._config.playback.playsinline === 'boolean') {
         this.playsinline = this._config.playback.playsinline;
       }
+      if (typeof this._config.playback.textStyle === "object") {
+        this.textStyle = Utils.Object.mergeDeep(new TextStyle(), this._config.playback.textStyle);
```
OrenMe: Player should know only TextStyle, if someone serializes class data outside of player he should recreate the TextStyle there and set it as player.textStyle = textStyle

src/player.js
```diff
@@ -794,6 +794,7 @@ export default class Player extends FakeEventTarget {
         sheet.insertRule(`#${this._playerId} .${SUBTITLES_CLASS_NAME} > div > div > div { ${style.toCSS()} }`, 0);
       }
       this._textStyle = style;
+      this.dispatchEvent(new FakeEvent(CustomEvents.TEXT_STYLE_CHANGED, {textStyle: this._textStyle.clone()}));
```
dan-ziv: Change the payload to:
` {textStyle: this.textStyle}`
it's already cloned the object,
so why access the private and clone it by yourself?

src/player.js

src/player.js
yairans: removed.

src/player.js
yairans: irrelevant. i've removed the payload. taken from `textStyle` getter.

dan-ziv:
@yairans 
approved just fix tests (you need to add the new event to the list in events.spec)

OrenMe:
@yairans test is failing, please fix

OrenMe:
@dan-ziv it's not clear to me what the test, which is now failing, does?! it just checks that the test array equals the actual player array. I think it holds no value and we can just remove it?

PR #144, opened 2017-10-22
feat(autoplay): manage autoplay promise and allowMutedAutoPlay flag

dan-ziv:

When the playkit module is loaded, we run statically test to check the video capabilities on the runtime environment. Those results (which are promise based) will be stored in each engine (currently only on Html5 engine) for later use. When autoplay is configured as true, before calling play(), we will check if the runtime env can autoplay: if yes, continue regular (call play internally). if not, continue according to the allowMutedAutoPlay flag.

Note: UI adjustments needs to be done in playkit-js-ui.

### CheckLists

- [x] Docs have been updated


src/player.js
```diff
@@ -1032,38 +1043,45 @@ export default class Player extends FakeEventTarget {
...
+      if (this._canPreloadAuto()) {
+        this.load();
       }
-      if (this._canAutoPlay()) {
-        this.play();
+      if (this._config.playback.autoplay === true) {
+        if (this.muted) {
+          this.play();
+        } else {
+          this._manageAutoPlay();
```
OrenMe: change to `_handleAutoPlay`

src/engines/html5/html5.js
```diff
@@ -109,6 +132,14 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   }
 
   /**
+   * Gets the result from the canAutoPlay test.
+   * @returns {Promise<boolean>} - The can auto play result.
+   */
+  canAutoPlay(): Promise<boolean> {
+    return Html5.canAutoPlay();
```
OrenMe: why do you need instance method as well as static?

src/player.js
```diff
@@ -112,6 +112,17 @@ export default class Player extends FakeEventTarget {
...
   static _engines: Array<typeof IEngine> = [Html5];
+
+  /**
+   * Runs the engines tests for autoplay capabilities.
+   * @static
+   * @public
+   * @returns {void}
+   */
+  static testAutoPlayCapabilities(): void {
+    Player._engines.forEach(Engine => Engine.canAutoPlay());
```
OrenMe: for future proof I would expose a capabilities object that can be quarried from player.
It can expose all player capabilities and by nature would be a promise.all that awaits all capabilities metrics to be finished.
Exposing each capability on it's own will result in API bloating

src/engines/html5/html5.js
dan-ziv: Because I don't know which engine will be chosen at run time. We can alternativly access this._engine.constructor.canAutoPlay() but I think it's ugly. Any suggestion?

src/player.js
dan-ziv: of course, while implemented this I cannot think of other capability for generic impl :)

src/player.js

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+ >If sets to `true` and the runtime browser blocks autoplay, 
+ >autoplay will start muted, until any user gesture 
+ >will be executed on the player, which will then unmute it. If sets to `false` and the runtime browser blocks autoplay, playback will not start automatically.
+
+ The following matrix summarise the results for configure `playback.allowMutedAutoPlay` when `playback.autoplay=true`:
+
+| Browser Policy          | `allowMutedAutoPlay=true` |` allowMutedAutoPlay=false`
+| ----------------- | ----------------- | ------ |
+| Blocks autoplay  | muted autoplay | no autoplay (user gesture is required) |
+| Allows autoplay  | autoplay with/without sound according to the player configuration | autoplay with/without sound according to the player configuration
```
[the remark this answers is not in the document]
dan-ziv: Ok, I will wait for Anna review and fix all together

src/player.js
OrenMe: ok, but we can already generalize the API to be more general like getCapabilities ()
which will return a promise that resolves all capabilities API calls.
This can be exposed in the Playkit API where you currently expose only autoplay explicitly

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
+## Managing Auto Play
+If you would like to start automatically playing your content, you can easily manipulating the player configuration and set the `playback.autoplay` value accordingly.
```
AnnaAleksandrowicz: If you would to start playing your content automatically, simply set the value of the Player configuration 'playback.autoplay' accordingly.

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+
+>### playback.autoplay
+>##### Type: `boolean`
+>##### Default: `false`
+>##### Description: Start playback automatically.
+>If `true`, playback will start automatically when sources will be given to the player. If `false`, a user gesture will be required to start playback.
```
AnnaAleksandrowicz: If set to `true`, playback will start automatically when sources will be provided to the Player. If set to `false`, a user action will be required to start playback.

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+};
+// No need to add additional code, 
+// playback will start automatically!
+var player = Playkit.loadPlayer(config);
+```
+
+>**_Note_**:
+> Some developers may 
+> simulate the autoplay behavior by calling `play()` method right after the creation of the player without manipulating the player configuration. We strongly recommend **not** doing it this way!
+>```js
```
AnnaAleksandrowicz: Some developers simulate the autoplay behavior by calling the `play()` method right after creating the Player without manipulating the player configuration. However, we strongly recommend **not** using this method!

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+>var config = {
+>  sources: {...}
+>};
+>var player = Playkit.loadPlayer(config);
+>// Bad practice!
+>player.play(); 
+>```
+> #### Why it's bad?
```
AnnaAleksandrowicz: Why not use this Method?

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+> When it's done as noted above, the player has no way to know whether the call to `play()` has done via user action or via API call, a fact that can cause unexpected behaviour on browsers which are not support autoplay.
```
AnnaAleksandrowicz: When you use the method described above, the Player has no way of knowing whether the call to 'play()' has been implemented via user action or via an API call. This can cause unexpected behavior in browsers that do not support autoplay.

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
+## Managing Auto Play
```
AnnaAleksandrowicz: Managing Autoplay on the Player

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+>var player = Playkit.loadPlayer(config);
+>// Bad practice!
+>player.play(); 
+>```
+> #### Why it's bad?
+> When it's done as noted above, the player has no way to know whether the call to `play()` has done via user action or via API call, a fact that can cause unexpected behaviour on browsers which are not support autoplay.
+### What if the browser does not allow autoplay?
+Unfortunately, on some platforms, web browsers does not allow to start playback automatically.
+The platforms that are blocks those attempts are mostly targeted to web browsers on mobile devices, but lately desktop browsers
+ are also start to align those restrictions in order to create unified behavior between platforms (currently Safari 11 and began of Chrome 64).
```
AnnaAleksandrowicz: On some platforms, web browsers do not allow to playback to begin automatically. The platforms that block attempts at playback are mostly targeted at web browsers on mobile devices; however, some desktop browsers have also started to align these restrictions to create a unified behavior between platforms (currently Safari 11 and beginning with Chrome 64).

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+ #### The Good News
+These restrictions apply only to automatic playback with sound, but not to automatic playback without sound.
+Our player can manage this logic by himself and to identify if the current environment does not support autoplay with sound. In that case, he will apply autoplay without sound.
+To use this feature, you'll need to set the `playback.allowMutedAutoPlay` accordingly to your desired behavior.
```
AnnaAleksandrowicz: The Kaltura Player can manage this logic on its own, and identify whether the current environment does not support autoplay with sound. In this case, the player will apply the autoplay with sound.

To use this feature, you'll need to set the `playback.allowMutedAutoPlay` accordingly to the desired behavior.

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+ >### playback.allowMutedAutoPlay
+ >##### Type: `boolean`
+ >##### Default: `true`
+ >##### Description: Start playback automatically without sound in case autoplay with sound is not allowed.
+ >If sets to `true` and the runtime browser blocks autoplay, 
+ >autoplay will start muted, until any user gesture 
+ >will be executed on the player, which will then unmute it. If sets to `false` and the runtime browser blocks autoplay, playback will not start automatically.
+
```
AnnaAleksandrowicz: >##### Description: Start playback automatically without sound if autoplay with sound is not allowed.
 + >If set to `true` and the run-time browser blocks autoplay, autoplay will begin muted until any user action is executed on the player, which will then unmute playback. If set to `false` and the run-time browser blocks autoplay, playback will not start automatically.

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+ The following matrix summarise the results for configure `playback.allowMutedAutoPlay` when `playback.autoplay=true`:
```
AnnaAleksandrowicz: The following matrix summarizes the results for configuring `playback.allowMutedAutoPlay` when `playback.autoplay=true`:

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+
+| Browser Policy          | `allowMutedAutoPlay=true` |` allowMutedAutoPlay=false`
+| ----------------- | ----------------- | ------ |
+| Blocks autoplay  | muted autoplay | no autoplay (user gesture is required) |
```
AnnaAleksandrowicz: | Blocks autoplay  | Muted autoplay | No autoplay (user action is required) |

docs/autoplay.md
```diff
@@ -0,0 +1,82 @@
...
+| Allows autoplay  | autoplay with/without sound according to the player configuration | autoplay with/without sound according to the player configuration
```
AnnaAleksandrowicz: | Allows autoplay  | Autoplay with/without sound according to the player configuration | Autoplay with/without sound according to the player configuration

docs/autoplay.md
dan-ziv: @AnnaAleksandrowicz 
Maybe my sentence wasn't clear so u didn't understood me.
The thing is that to use this method is fine and necessary - for example if you pause the player, the only way to resume playback is to call play.
What I meant here is not to use it as "fake autoplay" - we want developers to let us manage to autoplay internally in the player core.

docs/autoplay.md
dan-ziv: @AnnaAleksandrowicz  see my comment above

AnnaAleksandrowicz: @dan-ziv - sorry I wasn't commenting - i was suggesting to rename the heading ""Why not use this method?" Sorry if I wasn't clear!

src/player.js
```diff
@@ -1132,6 +1199,7 @@ export default class Player extends FakeEventTarget {
     this._activeTextCues = [];
     this._tracks = [];
     this._firstPlay = true;
+    this._firstPlayInCurrentSession = false;
```
OrenMe: @dan-ziv wouldn't this reset on change media?

dan-ziv: @OrenMe 
yes, to false - this is exactly what we want - the is not the first play in current session now

src/player.js
```diff
@@ -1078,6 +1078,10 @@ export default class Player extends FakeEventTarget {
       });
       this._eventManager.listen(this._engine, CustomEvents.TEXT_CUE_CHANGED, (event: FakeEvent) => this._onCueChange(event));
       this._eventManager.listen(this._engine, CustomEvents.ABR_MODE_CHANGED, (event: FakeEvent) => this.dispatchEvent(event));
+      this._eventManager.listen(this._engine, CustomEvents.AUTOPLAY_FAILED, (event: FakeEvent) => {
+        this.pause();
```
itaykinnrot: On Safari11 we'll get pause event from the browser too so maybe check the player state before sending the pause command?

PR #145, opened 2017-10-24
fix(fec-7334): the player state has not saved on change media

yairans:
on change media we are considering the player config but not to the current state of the following player attributes: 
* muted
* volume
* text language
* audio language
* playback rate

the solution is: 
* saving the state of the attributes above
* considering them (on change media) before player config  
 
### CheckLists

- [x] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1310,8 +1310,10 @@ export default class Player extends FakeEventTarget {
 
     this.hideTextTrack();
 
-    this._setDefaultTrack(TrackTypes.TEXT, this._playbackAttributesState.textLang, this._getLanguage(playbackConfig.textLanguage, activeTracks.text, TrackTypes.TEXT), offTextTrack);
-    this._setDefaultTrack(TrackTypes.AUDIO, this._playbackAttributesState.audioLang, playbackConfig.audioLanguage, activeTracks.audio);
+    let currentOrConfiguredTextLang = this._playbackAttributesState.textLang || this._getLanguage(playbackConfig.textLanguage, activeTracks.text, TrackTypes.TEXT);
+    let currentOrConfiguredAudioLang = this._playbackAttributesState.audioLang || playbackConfig.audioLanguage;
+    this._setDefaultTrack(TrackTypes.TEXT, currentOrConfiguredTextLang, offTextTrack);
+    this._setDefaultTrack(TrackTypes.AUDIO, currentOrConfiguredAudioLang, activeTracks.audio);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/player.js
```diff
@@ -1002,53 +1019,71 @@ export default class Player extends FakeEventTarget {
...
+        this._playbackAttributesState.TEXT_LANG = event.payload.selectedTextTrack.language;
         this._markActiveTrack(event.payload.selectedTextTrack);
         return this.dispatchEvent(event);
       });
       this._eventManager.listen(this._engine, CustomEvents.TEXT_CUE_CHANGED, (event: FakeEvent) => this._onCueChange(event));
       this._eventManager.listen(this._engine, CustomEvents.ABR_MODE_CHANGED, (event: FakeEvent) => this.dispatchEvent(event));
       this._eventManager.listen(this, Html5Events.PLAY, this._onPlay.bind(this));
       this._eventManager.listen(this, Html5Events.ENDED, this._onEnded.bind(this));
+      this._eventManager.listen(this, Html5Events.VOLUME_CHANGE, () => {
+        this._playbackAttributesState.MUTED = this.muted;
```
dan-ziv: split to MUTE_CHANGE

src/player.js
```diff
@@ -1104,6 +1139,9 @@ export default class Player extends FakeEventTarget {
       this._firstPlay = false;
       this.dispatchEvent(new FakeEvent(CustomEvents.FIRST_PLAY));
       this._posterManager.hide();
+      if (typeof this._playbackAttributesState.RATE === 'number') {
+        this.playbackRate = this._playbackAttributesState.RATE;
```
dan-ziv: place the update for playbackRate in ratechange event listener (need to add if we don't have this event)

yairans: @dan-ziv 
in `ratechange` i'm saving the rate value. here i'm using it.

src/player.js

src/player.js
dan-ziv: So why here?
We cannot assign it also in `_handlePlaybackOptions`?

yairans: no. you can change the rate only after loading/play

src/player.js
```diff
@@ -1270,8 +1310,8 @@ export default class Player extends FakeEventTarget {
 
     this.hideTextTrack();
 
-    this._setDefaultTrack(TrackTypes.TEXT, this._getLanguage(playbackConfig.textLanguage, activeTracks.text, TrackTypes.TEXT), offTextTrack);
-    this._setDefaultTrack(TrackTypes.AUDIO, playbackConfig.audioLanguage, activeTracks.audio);
+    this._setDefaultTrack(TrackTypes.TEXT, this._playbackAttributesState.TEXT_LANG, this._getLanguage(playbackConfig.textLanguage, activeTracks.text, TrackTypes.TEXT), offTextTrack);
```
OrenMe: @yairans `playbackConfig.textLanguage` is initial config and if there is `this._playbackAttributesState.TEXT_LANG` already it means that it replaced the initial config, so why pass both now? just pass this or that to the `_setDefaultTrack` method

src/player.js
```diff
@@ -226,6 +226,19 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _streamType: string;
+  /**
+   * The current playback attributes state
+   * @type {Object}
+   * @private
+   */
+  _playbackAttributesState: { [attribute: string]: any } = {
+    MUTED: undefined,
```
OrenMe: @yairans why use CAPITAL? it's not a const just an object map. please use lower case.

src/player.js
yairans: @OrenMe the `this._playbackAttributesState.TEXT_LANG` could be non exists in the new media. for this case we have to take the lang form config.

OrenMe: The original config can be non exist as well, it doesn’t matter let this case. The point is, the only way you have a config here is if it was previously set, but is is only one, either from app config, auto selected or user selected.

src/player.js

PR #147, opened 2017-10-25
fix(FEC-7106): captions on IE & edge

odedhutzler:

patch to fix the captions issue on ie and edge, until this issue is resolved. when seeking to a place without captions, the previous captions (if there were any) are still shown. this patch removes them.

### CheckLists

- [x] Docs have been updated


src/player.js
```diff
@@ -1045,6 +1051,25 @@ export default class Player extends FakeEventTarget {
...
 
   /**
+   * Handles the cue text removal issue, when seeking to a time without captions in IE \ edge the previous captions
+   * are not removed
+   * @returns {void}
+   * @private
+   */
+_removeCueTextPatch(): void {
+  const textCue = this._activeTextCues;
+  if (textCue.length > 0) {
```
OrenMe: `_activeTextCues` can be an array so need to iterate over all of it and throw out the out of bound cues and keep the ones in it

src/player.js
```diff
@@ -1097,6 +1103,23 @@ export default class Player extends FakeEventTarget {
...
+   * are not removed
+   * @returns {void}
+   * @private
+   */
+  _removeTextCuePatch(): void {
+    this._activeTextCues.forEach((textCue) => {
+      const cueEndTime = textCue._endTime;
+      const cueStartTime = textCue._startTime;
+      const currTime = this.currentTime;
+      if (currTime > cueEndTime || currTime < cueStartTime) {
```
OrenMe: you need to prepare the array you send to `processCues` before calling it as it is not aware of time, just display.
Calling it like this will cause only the last cue in the array to appear

PR #155, opened 2017-10-31
fix(FEC-7266): reposition & resize captions when changing to/from full screen

odedhutzler:

Changing the single full screen event to fire also an event when it ends (its a hack, we wait 500ms and fire the event because currently browsers does not support a "finished render full screen" event).

After the full screen event, we reposition and resize the captions.

+ adding tests.

### CheckLists

- [x] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -33,13 +33,6 @@ import './assets/style.css'
 const CONTAINER_CLASS_NAME: string = 'playkit-container';
 
 /**
- * The toggle fullscreen rendering timeout value
- * @type {number}
- * @const
- */
-const FULLSCREEN_RENDERING_TIMEOUT: number = 1000;
-
-/**
  /**
  * The player poster class name.
  * @type {string}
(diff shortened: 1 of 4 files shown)


src/player.js
```diff
@@ -1157,10 +1157,34 @@ export default class Player extends FakeEventTarget {
...
+   * @private
+   */
+  _resetTextCuesAndReposition(): void {
+    this._updateTextDisplay([]);
+    for (let i = 0; i < this._activeTextCues.length; i++) {
+      this._activeTextCues[i].hasBeenReset = true;
+    }
+
+    // waiting the browser to render the screen
+    setTimeout(() => {
```
OrenMe: save timeout id and handle cases when events come in consecutive manner, e.g. debounce it.
Also, move timeout to const value.

odedhutzler: @OrenMe i didn't debounce it - just ignored previous timeouts (the last one will be the one that will be executed - i think that's the behavior we want)

src/player.js
```diff
@@ -1157,7 +1158,34 @@ export default class Player extends FakeEventTarget {
...
+      this._activeTextCues[i].hasBeenReset = true;
+    }
+    // handling only the last reposition
+    if (this._repositionCuesTimeout) {
+      clearTimeout(this._repositionCuesTimeout);
     }
+    this._repositionCuesTimeout = setTimeout(() => {
+      processCues(window, this._activeTextCues, this._textDisplayEl);
+      this._repositionCuesTimeout = false;
+    }, 1000);
```
OrenMe: Just move this to a const value, REPOSITION_CUES_TIMEOUT and we are good

PR #156, opened 2017-11-02
feat: handle native selection of tracks

dan-ziv:

On iOS, when choosing from the native player (AVPlayer) audio track or text track, the player UI wasn't updated. 
This PR bring sync between the player UI and the native UI.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -75,11 +75,17 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
    */
   _progressiveSources: Array<Source>;
   /**
-   * The player tracks.
+   * The player text tracks.
    * @member {Array<Track>} - _playerTracks
    * @private
    */
-  _playerTracks: Array<Track>;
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -249,15 +258,15 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+   * @function _parseTracks
+   * @returns {void}
    * @private
    */
-  _getParsedTracks(): Array<Track> {
-    let videoTracks = this._getParsedVideoTracks();
-    let audioTracks = this._getParsedAudioTracks();
-    let textTracks = this._getParsedTextTracks();
-    return videoTracks.concat(audioTracks).concat(textTracks);
+  _parseTracks(): void {
```
OrenMe: why not return tracks from here? we do the same in HLS and DASH adapters so why change convention?

dan-ziv: done

PR #158, opened 2017-11-07
fix(FEC-7355): don't show poster when autoplaying is on

odedhutzler:

When autoplay is on, the poster might be shown for few miliseconds, which looks buggy.
If autoplay is on & the browser is capable of doing it, I am hiding the poster element .

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -346,8 +346,6 @@ export default class Player extends FakeEventTarget {
       }
       if (this._selectEngineByPriority()) {
         this._appendEngineEl();
-        this._posterManager.setSrc(this._config.metadata.poster);
-        this._posterManager.show();
         this._attachMedia();
         this._handlePlaybackOptions();
         this._handleAutoPlay();
@@ -1271,17 +1269,16 @@ export default class Player extends FakeEventTarget {
           .then((capabilities) => {
             if (capabilities.autoplay) {
               Player._logger.debug("Start autoplay");
-              this._posterManager.hide();
               this.play();
             } else {
               if (allowMutedAutoPlay) {
                 Player._logger.debug("Fallback to muted autoplay");
                 this.muted = true;
-                this._posterManager.hide();
                 this.play();
                 this.dispatchEvent(new FakeEvent(CustomEvents.FALLBACK_TO_MUTED_AUTOPLAY));
               } else {
                 Player._logger.warn("Autoplay failed, pause player");
+                this._posterManager.setAndShow(this._config.metadata.poster);
                 this.load();
                 this.ready().then(() => this.pause());
                 this.dispatchEvent(new FakeEvent(CustomEvents.AUTOPLAY_FAILED));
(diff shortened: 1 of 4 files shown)


src/player.js
```diff
@@ -320,8 +320,8 @@ export default class Player extends FakeEventTarget {
     this._textStyle = new TextStyle();
     this._createReadyPromise();
     this._createPlayerContainer();
-    this._appendPosterEl();
     this.configure(config);
+    (this.config.playback && !this.config.playback.autoplay) && this._appendPosterEl();
```
OrenMe: This is the poster DOM placeholder, you want to avoid setting the poster source, not the entire DOM element

OrenMe:
@odedhutzler if change media seems like a related change, IMO it is, then do it in same PR

src/player.js
dan-ziv: @OrenMe you don't evenn want to avoid setting the poster source but only avoid to show it (i.e. posterManager.show())

dan-ziv: @odedhutzler you can't rely on `config.playback.autoplay` in this manner since it's can be `true` but the browser disallow autoplay

src/player.js
```diff
@@ -346,8 +346,6 @@ export default class Player extends FakeEventTarget {
       }
       if (this._selectEngineByPriority()) {
         this._appendEngineEl();
-        this._posterManager.setSrc(this._config.metadata.poster);
```
dan-ziv: You want to set the source anyway, whether if it will be hidden or showing later.

src/utils/poster-manager.js
```diff
@@ -76,6 +76,17 @@ class PosterManager {
   }
 
   /**
+   * set src & Show the poster image
+   * @public
+   * @param {string} posterUrl - the poster image URL
+   * @returns {void}
+   */
+  setAndShow(posterUrl: ?string): void {
```
dan-ziv: First, this method is redundant (see my comment above)
Second, the name of it is too generic - set and show what?

src/player.js
```diff
@@ -1280,13 +1278,16 @@ export default class Player extends FakeEventTarget {
                 this.dispatchEvent(new FakeEvent(CustomEvents.FALLBACK_TO_MUTED_AUTOPLAY));
               } else {
                 Player._logger.warn("Autoplay failed, pause player");
+                this._posterManager.setAndShow(this._config.metadata.poster);
```
dan-ziv: call only ` this._posterManager.show()`

src/player.js
```diff
@@ -1280,13 +1278,16 @@ export default class Player extends FakeEventTarget {
...
                 this.load();
                 this.ready().then(() => this.pause());
                 this.dispatchEvent(new FakeEvent(CustomEvents.AUTOPLAY_FAILED));
               }
             }
           });
       }
+    } else {
+      this._posterManager.setAndShow(this._config.metadata.poster);
```
dan-ziv: call only ` this._posterManager.show()`

PR #161, opened 2017-11-11
feat: expose logger API

OrenMe:

set to ERROR level by default
expose API to set and get log level and get log level enums
expose convenience method to set logger level via global window param window.PLAYKIT_LOG_LEVEL
update to JS-Logger V1.4

### CheckLists

- [ ] Docs have been updated


src/utils/logger.js
```diff
@@ -10,20 +10,38 @@ const LOG_LEVEL: { [level: string]: Object } = {
...
+ * @returns {void}
+ */
+function setLogLevel(level: Object, name?: string): void{
+  getLogger(name).setLevel(level);
+}
 
-export default lf;
-export {LOG_LEVEL};
+export default getLogger;
+export {LOG_LEVEL, getLogLevel, setLogLevel};
```
dan-ziv: change the var name `LOG_LEVEL` to `LogLevel` since it's an enum and then we can export it regularly without use the `as` syntax of flow in other files

sug.txt
```diff
@@ -0,0 +1,107 @@
+/Users/OrenMe/repos/playkit-js/src/player.js
```
dan-ziv: what this file for?

text.md
```diff
@@ -0,0 +1,16 @@
...
+* Audio tracks - enable selecting alternative audio for a video
+* Text tracks - enable selecting different caption or subtitle text for a video
+
+All track types are represented in player UI and may be selected by user.
+In addition tracks can also be set via API.
+```javascript
+// create a player
+var player = KalturaPlayer.setup("player-placeholder", config);
+
+var textplayer.getTracks(player.Track.TEXT);
```
dan-ziv: did u mean
`var textTracks = player.getTracks(player.Track.TEXT);`
?

text.md
```diff
@@ -0,0 +1,16 @@
...
+player.selectTrack(track)
```
dan-ziv: where is `track`?

src/player.js
```diff
@@ -322,6 +322,9 @@ export default class Player extends FakeEventTarget {
     this._createPlayerContainer();
     this._appendPosterEl();
     this.configure(config);
+    if (config.logLevel && this.LogLevel[config.logLevel]){
```
dan-ziv: Why can't we do this in the `loadPlayer` factory function of playkit.js before the initialization of the player?
in configure we already write logs and if we are doing it after it won't affect this, no?

dan-ziv:
@OrenMe 
What are
cov.json
and 
sug.txt
?

src/player.js
dan-ziv: I see now that we are using this for the check, but we can define fallback level in such cases and avoid this

src/utils/logger.js
OrenMe: @dan-ziv it's either an enum or not.
If it's an enum we should keep the convention as we do with other enums, like:
`import {PLAYER_EVENTS as PlayerEvents, HTML5_EVENTS as Html5Events, CUSTOM_EVENTS as CustomEvents} from './event/events'
`

src/player.js
OrenMe: If you refer to the version and name printing in index.js, we don't want to print them and in any case these printouts are in load time before CTOR is called, and setting the config level can be done only in init time

dan-ziv: No, I'm not referring to the version and name printouts.
I'm mean that you set the log level too late - after the configure method.
The configure method flow will print logs and seems like those logs won't have the correct log level. 
Don't you think the correct place to handle the log level is before creating the player? in the loadPlayer method?

src/utils/logger.js
dan-ziv: Those conventions are wrong and already fixed in my player typing PR which you all ignore of reviewing :)
From what I see in all the lead projects the correct conventions for enums should be lead with capitals (i.e. MyEnum) and constants should be all written in capital letters (i.e. MY_CONST)

OrenMe: fixed

src/player.js
OrenMe: No, the static API should remain simple IMHO.
Constructor is where everything should be setup, and in the case of logs, this line should just some before others if it's more important.
Fixed by moving to be first in line

test/src/player.spec.js
```diff
@@ -2672,3 +2671,23 @@ describe('_resetTextCuesAndReposition', function () {
   });
 });
 
+describe('logger', ()=>{
```
dan-ziv: prettify the code in this file

src/playkit.js
```diff
@@ -21,8 +21,8 @@ declare var __VERSION__: string;
 declare var __NAME__: string;
 declare var __PACKAGE_URL__: string;
 
-LoggerFactory.getLogger().log(`%c ${__NAME__} ${__VERSION__}`, "color: #98ff98;  font-size: large");
-LoggerFactory.getLogger().log(`%c For more details see ${__PACKAGE_URL__}`, "color: #98ff98;");
+getLogger().log(`%c ${__NAME__} ${__VERSION__}`, "color: #98ff98;  font-size: large");
```
dan-ziv: Lets delete those logs... ?

src/player.js
```diff
@@ -306,6 +306,9 @@ export default class Player extends FakeEventTarget {
    */
   constructor(config: Object = {}) {
     super();
+    if (config.logLevel && LogLevel[config.logLevel]){
```
dan-ziv: prettify

dan-ziv:
@OrenMe approved just fix codeclimate and travis (same issue)

PR #170, opened 2017-12-04
feat: add stream metadata event(supported on hls for now)

itaykinnrot:
Add stream metadata event(id3tag:-) 

### CheckLists

- [x] Docs have been updated


src/engines/html5/html5.js
```diff
@@ -182,6 +188,11 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   attach(): void {
     Object.keys(Html5Events).forEach((html5Event) => {
       this._eventManager.listen(this._el, Html5Events[html5Event], () => {
+        if (Html5Events[html5Event] == Html5Events.PLAYING &&
+          !this._registeredMetadataEvent &&
+          this._config.playback.registerMetadataTrackEvent){
+          setTimeout(()=>{this._registerMetadata()},1000);
```
dan-ziv: @itaykinnrot why the timeout?

src/engines/html5/html5.js
```diff
@@ -182,6 +188,11 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   attach(): void {
     Object.keys(Html5Events).forEach((html5Event) => {
       this._eventManager.listen(this._el, Html5Events[html5Event], () => {
+        if (Html5Events[html5Event] == Html5Events.PLAYING &&
+          !this._registeredMetadataEvent &&
+          this._config.playback.registerMetadataTrackEvent){
```
dan-ziv: @itaykinnrot  why we need this configurable?

src/engines/html5/html5.js
```diff
@@ -880,5 +892,32 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+          this._registeredMetadataEvent = true;
+          tracks[track].addEventListener("cuechange", (evt) => {
+            let activeCues =[];
+            for (let cue in evt.currentTarget.activeCues){
+              if (evt.currentTarget.activeCues.hasOwnProperty(cue)) {
+                let currentCue = evt.currentTarget.activeCues[cue];
+                activeCues.push(currentCue)
+              }
+            }
+            this.dispatchEvent(new FakeEvent(CustomEvents.STREAM_METADATA,{activeCues}));
```
dan-ziv: need to add key to the payload object

src/engines/html5/html5.js
```diff
@@ -880,5 +892,32 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+   * @private
+   */
+  _registerMetadata(): void {
+    const tracks = this._el.textTracks;
+    for (let track in tracks) {
+      if (tracks.hasOwnProperty(track)){
+        if (tracks[track].kind == "metadata") {
+          tracks[track].mode = "showing";
+          this._registeredMetadataEvent = true;
+          tracks[track].addEventListener("cuechange", (evt) => {
```
dan-ziv: @itaykinnrot we are already listen to this event and running this logic in the engine, I'm wondering if we can reuse the existing listener to your needs?

src/engines/html5/html5.js
```diff
@@ -54,6 +54,12 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @private
    */
   _canLoadMediaSourceAdapterPromise: Promise<*>;
+  /**
+   * Promise to indicate when a media source adapter can be loaded.
+   * @type {Promise<*>}
+   * @private
+   */
+  _registeredMetadataEvent: boolean;
```
dan-ziv: Do we really need this flag? can we rid of it in some way?

dan-ziv:
@itaykinnrot please beautify your code

src/engines/html5/html5.js
itaykinnrot: We want to enable it only if a plugin is requesting this info - since sometimes it can cause issues in browsers (safari :-( )

src/engines/html5/html5.js
itaykinnrot: On safari the metadata track is visible only in async call after playing - not sure why

src/engines/html5/html5.js
itaykinnrot: No since the other event is for the real text - this one is for metadata -- id3 tags

src/engines/html5/html5.js
itaykinnrot: ?

src/engines/html5/html5.js
itaykinnrot: Yes - since we want to make sure we try to find this track in every playback try

src/engines/html5/html5.js
dan-ziv: {activeCues: activeCues}

src/engines/html5/html5.js
dan-ziv: So need to reset it on change media?

itaykinnrot:
@dan-ziv  upadte the docs on the configuration branch and added a test

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -717,7 +717,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
    */
   _disableTextTracks(): void {
     let textTracks = this._videoElement.textTracks;
-    if (textTracks) {
+    if (textTracks && textTracks.kind !="metadata") {
```
OrenMe: should be `["captions", "subtitles"].includes(textTracks.kind);` as there are [other kinds as well](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track).
And also, we should add textrack kinds as an ENUM in the `./tracks` folder and import it here so we can use it as
```javascript
import {TEXT_TRACK_KINDS} from ...
TEXT_TRACK_KINDS.CAPTIONS;
TEXT_TRACK_KINDS.SUBTITLES;
```

so we can avoid some previously notorious bugs with caption vs captions string errors....

OrenMe: do we need to handle this in hls and dash adapters as well?

src/engines/html5/html5.js
OrenMe: @itaykinnrot if we need to do this let's move it to it's own event listener:
```javascript
this._eventManager.listen(this._el, Html5Events.PLAYING, () => this. _shouldAddMetadataEvent);
```

and move the logic if to add it or not to the handler itself

src/engines/html5/html5.js
OrenMe: @itaykinnrot I think we should open a bug to apple on this

src/engines/html5/html5.js
```diff
@@ -881,5 +893,32 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+  /**
+   * Register the metadata track event
+   * @returns {void}
+   * @private
+   */
+  _registerMetadata(): void {
+    const tracks = this._el.textTracks;
+    for (let track in tracks) {
+      if (tracks.hasOwnProperty(track)){
+        if (tracks[track].kind == "metadata") {
```
OrenMe: use KIND enum

OrenMe:
Also, would generally need to add protection in every place we iterate vid.textTracks for tracks of captions and subtitles kind, that they are of that kind, same goes for metadata

src/engines/html5/media-source/adapters/native-adapter.js
itaykinnrot: Since both are using the video tag - this cover all of them.

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -572,7 +573,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
   selectTextTrack(textTrack: PKTextTrack): void {
     const textTracks = this._videoElement.textTracks;
     if ((textTrack instanceof PKTextTrack)
-      && (textTrack.kind === 'subtitles' || textTrack.kind === 'captions')
+      && (textTrack.kind === TRACK_KIND.SUBTITLES || textTrack.kind === TRACK_KIND.CAPTIONS)
```
OrenMe: you already do `[TRACK_KIND.CAPTIONS, TRACK_KIND.SUBTITLES].includes(textTracks.kind)` a few lines below, so let's use it here as well.

OrenMe:
done as part of DAI

PR #175, opened 2018-01-08
fix(FEC-7371): videos are not playing when autoplay=true on old browsers

odedhutzler:

Older browsers (safari 9 e.g.) does not return a promise when  videoElement.'play()'

Added a timeout and error listener to see if the play() call failed, added 'playing' event listener to see if the player started to play.

Added muted autoplay capability check - because these browsers does not support this as well.

### CheckLists

- [ ] Docs have been updated


OrenMe:
LGTM, @dan-ziv , @yairans ?

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -15,12 +22,35 @@ export default class Html5AutoPlayCapability implements ICapability {
...
       Html5AutoPlayCapability._promise = Promise.reject();
     }
   }
 
+  /***
+   * Runs the test for autoplay capability in older browsers, where play() does not return a promise
+   * @static
+   * @returns {Promise<*>} - return a resolve/reject upon the success of playing
+   */
+  static _oldBrowsersCapability(): Promise<*> {
```
dan-ziv: Why do we need this check? please explain . 
Anyway this is not capability is not accurate

odedhutzler: Old browsers (such as safari 9) does not return a Promise when activating play(), it returns 'undefined'. In this case (autoplay), it does nothing (and not throwing an error or something..)

So the only way to see that the video start playing is with timeout (any other idea?)

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,35 +1,111 @@
...
 
+  /***
+   * Runs the test for autoplay capability in older browsers, where play() does not return a promise
+   * @static
+   * @param {HTMLVideoElement} element - video element to test
+   * @returns {Promise<*>} - return a resolve/reject upon the success of playing
+   */
+  static _oldBrowsersCapability(element: HTMLVideoElement): Promise<*> {
+    return new Promise((resolve, reject) => {
+      let supported = setTimeout(() => {
```
OrenMe: `const supported`

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,35 +1,111 @@
...
 import * as Utils from '../../../utils/util'
 import * as EncodingSources from '../../../assets/encoding-sources.json'
+import {HTML5_EVENTS as Html5Events} from '../../../event/events'
 
-export default class Html5AutoPlayCapability implements ICapability {
+/**
+ * The time (miliseconds) we wait that the playing event will be fired
+ * @const
+ */
+const playTimeoutBuffer = 1500;
```
OrenMe: internal members/const should have `_` prefix

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,35 +1,111 @@
...
+
+  /***
+   * Runs test for autoplay capability.
+   * @static
+   * @returns {void}
+   */
+  static _runAutoplayCapabilty(): void {
+    try {
+      Html5AutoPlayCapabilities._vid.src = EncodingSources.Base64Mp4Source;
+      Html5AutoPlayCapabilities._autoPlayPromise = Html5AutoPlayCapabilities._vid.play() || Html5AutoPlayCapabilities._oldBrowsersCapability(Html5AutoPlayCapabilities._vid);
```
OrenMe: @odedhutzler need to verify that this won't cause audible playback

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,35 +1,111 @@
...
+   * Gets test result for autoplay & muted autoplay capabilities.
    * @returns {Promise<CapabilityResult>} - The result object for autoplay capability.
    * @static
    * @public
    */
   static getCapability(): Promise<CapabilityResult> {
-    return Html5AutoPlayCapability._promise
-      .then(() => ({autoplay: true}))
-      .catch(() => ({autoplay: false}));
+    const promiseArr = [Html5AutoPlayCapabilities._autoPlayPromise, Html5AutoPlayCapabilities._mutedAutoPlayPromise];
```
OrenMe: I think this better be cascading check, e.g. first try autoplay and only then muted autoplay cause if autoplay is supported we don't need to try muted autoplay

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,35 +1,111 @@
...
+
+  /***
+   * act as a proxy to the promises. Promise.all is rejected if one of the promises is rejected, and the other promises
+   * data is lost, so here, we always resolve the promise, with status false.
+   * @param {Promise<*>} promise - one of the capabilities result.
+   * @returns {Promise<*>} - promise after manipulation.
+   * @static
+   * @public
+   */
+  static _reflect(promise: Promise<*>): Promise<*> {
```
OrenMe: reflect might not be an informative method name

src/engines/html5/capabilities/html5-autoplay.js
odedhutzler: @OrenMe  how can it? the browser won't allow it (?)

src/engines/html5/capabilities/html5-autoplay.js
odedhutzler: K, will change it.

src/engines/html5/capabilities/html5-autoplay.js
OrenMe: if browser allows it this test will cause sound

odedhutzler: @OrenMe so we need another method to check this capability, that's the way it checks autoplay in the current version of the player.

PR #185, opened 2018-01-30
fix(FEC-7872): need to click one more time on player area in order to start playback

dan-ziv:

Problem:
Player V3: Player stuck in case of multirequest is not fully loaded (need to click one more time on player area in order to start playback)

Solution:
In case play() has been called before engine is selected wait for SOURCE_SELECTED event and then trigger play() again.

Now we can perform safely calls to play() even if its too early.

### CheckLists

- [ ] Docs have been updated


OrenMe:
Need to check this with Safari or Chrome66(where autoplay restriction should be active) as it looks like autoplay procedure

PR #189, opened 2018-02-08
fix(FEC-7896): endless spinner appears right after changing audio track in IE and Edge

odedhutzler:

Handling a situation after audio track is changed, only in IE and Edge.
If a waiting event is sent, we send playing event. 

### CheckLists

- [ ] Docs have been updated


src/engines/html5/html5.js
```diff
@@ -200,6 +202,23 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+   * This align Edge and IE behaviour to other browsers. When an audio track changed in IE & Edge, they trigger
+   * waiting event but not playing event.
+   * @returns {void}
+   */
+  _handleWaitingEventOnAudioTrack(): void {
+    const affectedBrowsers = ['IE', 'Edge'];
+    if (affectedBrowsers.includes(Env.browser.name)) {
+      this._eventManager.listen(this._mediaSourceAdapter, CustomEvents.AUDIO_TRACK_CHANGED, () => {
+        this._eventManager.listenOnce(this._el, Html5Events.WAITING, () => {
+          this.dispatchEvent(new FakeEvent(Html5Events.PLAYING));
```
itaykinnrot: @odedhutzler  2 issues that we need to fix here:
1. we saw this issue only with hls right? so lets move the code to hls
2. if we change the audio and really going into a buffer the UI will not show it right?

PR #197, opened 2018-02-21
fix(FEC-7872): Player stuck in case of multirequest is not fully loaded

yairans:
* Solution for Android Chrome which requires user gesture on the specific video element (part of https://github.com/kaltura/playkit-js/pull/193)
* Do not wait to `sourceselected` event before ad playing, play ad immediately, see https://github.com/kaltura/playkit-js-ima/pull/45  

### CheckLists

- [x] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -440,11 +440,7 @@ export default class Player extends FakeEventTarget {
     } else if (this._loadingMedia) {
       // load media requested but the response is delayed
       Player._prepareVideoElement();
-      if (this.config.plugins.ima) {
-        this._playbackMiddleware.play(() => this._playAfterAsyncAds());
-      } else {
-        this._eventManager.listenOnce(this, CustomEventType.SOURCE_SELECTED, () => this.play());
-      }
+      this._playbackMiddleware.play(() => this._playAfterAsyncMiddleware());
     } else {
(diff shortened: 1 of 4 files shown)


src/player.js
```diff
@@ -436,14 +436,17 @@ export default class Player extends FakeEventTarget {
    */
   play(): void {
     if (this._engine) {
-      this._playbackMiddleware.play(this._play.bind(this));
+      this._playbackMiddleware.play(() => this._play());
     } else if (this._loadingMedia) {
       // load media requested but the response is delayed
       Player._prepareVideoElement();
-      this._eventManager.listenOnce(this, CustomEventType.SOURCE_SELECTED, () => this.play());
+      if (this.config.plugins.ima) {
```
OrenMe: Make this condition more generic and add it as a generic config.
The application can decide this, in this case if we have IMA or in future another ad plugin, we can set it in Kaltura-Player, or application that uses Kaltura player can even force this condition.

OrenMe:
@yairans please fix title and description

src/player.js
OrenMe: let's make this callback name also more generic, as any middleware chain might want to use this. `_playAfterAsyncMiddleware` sounds short :satisfied: and concise

yairans:
@OrenMe  actually no any condition is needed, see my last commit

PR #199, opened 2018-02-26
fix(FEC-7939): 'off' option appears twice after change media

yairans:
Raising a flag when the `load` process is starting and blocking another `load` call

### CheckLists

- [x] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -427,13 +427,14 @@ export default class Player extends FakeEventTarget {
       this._loading = true;
       let startTime = this._config.playback.startTime;
       this._engine.load(startTime).then((data) => {
-        this._loading = false;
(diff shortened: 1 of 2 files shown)


dan-ziv:
@yairans  remove 'only' from tests

src/player.js
```diff
@@ -332,13 +332,18 @@ export default class Player extends FakeEventTarget {
...
    * @type {boolean}
    * @private
    */
   _loadingMedia: boolean;
+  /**
+   * Whether the player is loading a source.
+   * @type {boolean}
+   * @private
+   */
+  _loading: boolean;
```
OrenMe: loading is to generic term, please change to engineLoading

PR #200, opened 2018-02-26
fix(FEC-7371): video fails to play on old browsers or browsers with data saver mode on

dan-ziv:

If autoplay capability test fails, try to run muted autoplay and add it to the capability result. 
This will fix also the error message we had in the console if the autoplay failed.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/capabilities/html5-autoplay.js
+++ b/src/engines/html5/capabilities/html5-autoplay.js
@@ -8,7 +8,7 @@ const WAIT_TIME: number = 500;
 const testVideoElement: HTMLVideoElement = Utils.Dom.createElement('video');
 testVideoElement.src = EncodingSources.Base64Mp4Source;
 // For iOS devices needs to turn the playsinline attribute on
-testVideoElement.setAttribute('playsinline', 'playsinline');
+testVideoElement.setAttribute('playsinline', '');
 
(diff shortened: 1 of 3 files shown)


src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,10 +1,17 @@
...
 import * as Utils from '../../../utils/util'
 import * as EncodingSources from '../../../assets/encoding-sources.json'
+import {Html5EventType} from '../../../event/event-type'
+
+const WAIT_TIME: number = 500;
+
+const testVideoElement: HTMLVideoElement = Utils.Dom.createElement('video');
+testVideoElement.src = EncodingSources.Base64Mp4Source;
+// For iOS devices needs to turn the playsinline attribute on
+testVideoElement.setAttribute('playsinline', 'playsinline');
```
OrenMe: playsinline doesn't have a value, just add
`testVideoElement.setAttribute('playsinline', '');`

OrenMe: we do the same in html5 engine

dan-ziv: playsinline is a boolean attribute, so, any value turns it on, using the name of the attribute is an often used convention

PR #205, opened 2018-03-26
feat(FEC-8079): add reset API to player and use it for change media

dan-ziv:

* Add reset API to player
* Add reset API to engine
* remove src getter of native adapter and use the same implementation in the base adapter.
* Add 4 new events: reset started/ended, destroy started/ended

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/event/event-type.js
+++ b/src/event/event-type.js
@@ -96,79 +96,95 @@ const Html5EventType: EventTypes = {
 
 const CustomEventType: EventTypes = {
   /**
-   * Fires when the player enters fullscreen
+   * Fires when the player starts reset operation.
+   */
+  RESET_STARTED: 'resetstarted',
+  /**
+   * Fires when the player ends reset operation.
+   */
+  RESET_ENDED: 'resetended',
+  /**
+   * Fires when the player starts destroy operation.
+   */
+  DESTROY_STARTED: 'destroystarted',
+  /**
+   * Fires when the player ends destroy operation.
+   */
+  DESTROY_ENDED: 'destroyended',
+  /**
+   * Fires when the player enters fullscreen.
    */
   ENTER_FULLSCREEN: 'enterfullscreen',
   /**
-   * Fires when the player exits fullscreen
+   * Fires when the player exits fullscreen.
    */
   EXIT_FULLSCREEN: 'exitfullscreen',
   /**
-   * Fires when the player received a request to enter fullscreen
+   * Fires when the player received a request to enter fullscreen.
    */
   REQUESTED_ENTER_FULLSCREEN: 'requestedenterfullscreen',
   /**
-   * Fires when the player received a request to exit fullscreen
+   * Fires when the player received a request to exit fullscreen.
    */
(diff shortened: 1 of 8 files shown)


src/player.js
```diff
@@ -356,6 +367,11 @@ export default class Player extends FakeEventTarget {
     this._firstPlay = true;
     this._fullscreen = false;
     this._firstPlayInCurrentSession = true;
+    this._repositionCuesTimeout = false;
+    this._loadingMedia = false;
```
odedhutzler: hey, whats the the difference between _loading & _loadingMedia?

src/player.js
```diff
@@ -486,12 +502,42 @@ export default class Player extends FakeEventTarget {
...
+   */
+  reset(): void {
+    if (this._reset) return;
+    this.dispatchEvent(new FakeEvent(CustomEventType.RESET_STARTED));
+    this.pause();
+    this._eventManager.removeAll();
+    this._createReadyPromise();
+    this._activeTextCues = [];
+    this._updateTextDisplay([]);
+    this._tracks = [];
```
OrenMe: I think we can move the state attributes to an internal method and call them from reset and destroy as all of them appear in both methods.

src/event/event-type.js
```diff
@@ -96,79 +96,95 @@ const Html5EventType: EventTypes = {
 
 const CustomEventType: EventTypes = {
   /**
-   * Fires when the player enters fullscreen
+   * Fires when the player starts reset operation.
+   */
+  RESET_STARTED: 'resetstarted',
```
OrenMe: @dan-ziv why would we need the `STARTED`|`ENDED` semantics?
In both operations you call the `reset`|`destroy' on all player managed objects, e.g. event manager, plugins etc. so everyone has ability to respond in time to the operation.
IMO I think it's enough to have `PLAYER_RESET` and `PLAYER_DESTROYED` events.
WDYT?

src/player.js
dan-ziv: Those flags are not part of this change but in general:
`loading` means that `load()` has been called on the player but the load promise isn't resolved yet, i.e while loading the engine. This is to prevent multiple load calls while a previous call is already processing. 
`loadingMedia` flag is a workaround that we should remove. Its means that `loadMedia()` has been called on the kaltura player but the provider response isn't resolved yet.

src/event/event-type.js
dan-ziv: I wanted to allow full lifecycle support for those operations (same as we did with change media).
when reset/destroy started the player still has its data (state, tracks, poster, source...), so if a certain app/component would like to save something or do some action (UI, routing, god knows what) before reseting/destroying this is the right time. When reset/destroy ended the player is all cleaned up and the component/app can respond accordingly. 
As far as the player is concerned, each will manage it the way he sees fit, but the player has to announce when he starts and when he finishes.

As for the plugins, I think we should do the same there and provide lifecycle methods for plugins:
`beforeReset`
`afterReset`
`beforeDestroy`
`afterDestroy`

src/player.js

PR #207, opened 2018-04-01
fix(FEC-8090): after disabling captions and toggling full screen the last caption frame is displayed

odedhutzler:

I emptied the _activeTextCues array when switching to 'off' track, so there won't be any captions to show.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -948,7 +948,6 @@ export default class Player extends FakeEventTarget {
       } else if (track instanceof TextTrack) {
         if (track.language === OFF) {
           this.hideTextTrack();
-          this._activeTextCues = [];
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -948,6 +948,7 @@ export default class Player extends FakeEventTarget {
       } else if (track instanceof TextTrack) {
         if (track.language === OFF) {
           this.hideTextTrack();
+          this._activeTextCues = [];
```
OrenMe: Let's move to the `hideTextTrack` method

OrenMe:
@odedhutzler LGTM, merge

PR #215, opened 2018-04-17
fix(FEC-7937): video first frame is displayed before pre-roll ad starts

dan-ziv:

Add black cover div which relays under the poster div and displayed as long as first play didn't happened.
Also, this refactor the way were handled the layers of the playkit container.
The correct layers priority now is (from strongest to weakest):
1. Ads/plugins
2. Poster
3. Black cover
4. Subtitles
5. Video frames

Which gives priority for each layer with no need for use of z-index.

This generated DOM for this order under playkit-container is:
```
<video>
<div class="playkit-subtitles">
<div class="playkit-black-cover">
<div class="playkit-poster">
<div class="playkit-ads-container">

```
### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1291,14 +1291,15 @@ export default class Player extends FakeEventTarget {
    * @returns {void}
    */
   _appendDomElements(): void {
+    // Append playkit-subtitles
     this._textDisplayEl = Utils.Dom.createElement("div");
     Utils.Dom.addClassName(this._textDisplayEl, SUBTITLES_CLASS_NAME);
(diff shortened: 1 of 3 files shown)


src/player.js
```diff
@@ -628,6 +640,9 @@ export default class Player extends FakeEventTarget {
     this._resetStateFlags();
     this._playbackAttributesState = {};
     if (this._el) {
+      while (this._el.firstChild) {
```
OrenMe: Need to do `this._el = null` instead

PR #226, opened 2018-05-23
fix(FEC-8233): handle media & error recovering

odedhutzler:

added listener to `mediarevocered` event.
added `isPlaying` property that stores if the player is currently playing or paused.
when the recovered event is fired, we check the isPlaying property and then decide if to continue playing or doing nothing.

### CheckLists

- [ ] Docs have been updated


src/engines/html5/html5.js
```diff
@@ -55,6 +55,12 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @private
    */
   _canLoadMediaSourceAdapterPromise: Promise<*>;
+  /**
+   * Flag that indicates the status of the engine (playing or paused)
+   * @type {boolean}
+   * @private
+   */
+  _isPlaying: boolean = false;
```
dan-ziv: @odedhutzler  for playback states we can use the state manager and avoid of keep another state in the player itself.

odedhutzler: moved the logic to the player (and used the state manager). thanks!

PR #229, opened 2018-05-27
feat(FEC-8243): allow to disable specific plugins

odedhutzler:

added '_disabled' property in the base plugin. in can be set using the setter.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/plugin/base-plugin.js
+++ b/src/plugin/base-plugin.js
@@ -38,11 +38,6 @@ export default class BasePlugin implements IPlugin {
    */
   eventManager: EventManager;
   /**
-   * Weather it is disabled or not
-   * @member
-   */
-  static _disabled: boolean = false;
-  /**
    * The default configuration of the plugin.
    * Inherited plugins should override this property.
    * @type {Object}
@@ -104,26 +99,6 @@ export default class BasePlugin implements IPlugin {
       return Utils.Object.copyDeep(this.config[attr]);
     }
     return Utils.Object.copyDeep(this.config);
-  }
-
-  /**
-   * getter for _disabled property
-   * @returns {boolean} the disabled property
-   * @static
-   * @public
-   */
-  static get disabled(): boolean {
-    return this._disabled;
-  }
-
-  /**
-   * @param {boolean} disable - should we disable the plugin or not
-   * @returns {void}
-   * @static
-   * @public
-   */
-  static set disabled(disable: boolean): void{
-    this._disabled = disable;
   }
 
   /**
--- a/src/plugin/base-plugin.js
+++ b/src/plugin/base-plugin.js
@@ -38,6 +38,11 @@ export default class BasePlugin implements IPlugin {
    */
   eventManager: EventManager;
   /**
+   * Weather it is disabled or not
+   * @member
+   */
+  static _disabled: boolean = false;
+  /**
    * The default configuration of the plugin.
    * Inherited plugins should override this property.
    * @type {Object}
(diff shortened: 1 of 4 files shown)


src/plugin/base-plugin.js
```diff
@@ -37,6 +37,11 @@ export default class BasePlugin implements IPlugin {
    * @member
    */
   eventManager: EventManager;
+  /**
+   * Weather it is disabled or not
```
dan-ziv: whether...

src/plugin/base-plugin.js
```diff
@@ -37,6 +37,11 @@ export default class BasePlugin implements IPlugin {
...
+   * @member
+   */
+  static _disabled: boolean = false;
```
dan-ziv: why do we need to save the disabled property in the base plugin?
Should we just not load the plugin if it configured with `disabled: true`?

src/plugin/plugin-manager.js
```diff
@@ -82,15 +82,29 @@ export default class PluginManager {
       throw new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.RUNTIME_ERROR_NOT_REGISTERED_PLUGIN, name);
     }
     let pluginClass = PluginManager._registry.get(name);
-    if (pluginClass && pluginClass.isValid()) {
+    this._maybeDisablePlugin(pluginClass, config.disable);
```
dan-ziv: why do we need this method?

src/plugin/plugin-manager.js
```diff
@@ -82,15 +82,29 @@ export default class PluginManager {
...
+  /**
+   * sets the disabled property.
+   * @param {Function} pluginClass - the current plugin
+   * @param {boolean} disable - config
+   * @returns {void}
+   * @private
+   */
+  _maybeDisablePlugin(pluginClass: ?Function, disable: boolean): void {
+    if (typeof disable !== 'undefined' && pluginClass) {
+      pluginClass.disabled = disable;
```
dan-ziv: there's a reason that we save this state?

src/plugin/plugin-manager.js
```diff
@@ -82,15 +82,29 @@ export default class PluginManager {
       throw new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.RUNTIME_ERROR_NOT_REGISTERED_PLUGIN, name);
     }
     let pluginClass = PluginManager._registry.get(name);
-    if (pluginClass && pluginClass.isValid()) {
+    this._maybeDisablePlugin(pluginClass, config.disable);
+    if (pluginClass && pluginClass.isValid() && !pluginClass.disabled) {
```
dan-ziv: why not just
`if (pluginClass && pluginClass.isValid() && !config.disable) { ....` ?

src/plugin/plugin-manager.js
odedhutzler: i debated myself if it's an overkill. but i think it's good to save the state of the plugin and not to rely solely on the config. we might want to disable the plugin from within (e.g. disable analytic s on offline mode). another use case i see is when the config of the plugin is not reachable / complex to get.

OrenMe: @odedhutzler I think offline is something that might change across sessions while disable is deterministic so we can indeed simplify this and just check if plugin config has disable true and just not initiate it.
There's no added value in setting the statics of the class, as plugins in player are only used as instance on a player instance.
Think about it this way, if you update the statics of the class another player might be instantiated and static disabled of it will also be true, although it wasn't set like that in the setup of it.

It's enough to reduce this to a conditional check in the `load` method and call it a day.

odedhutzler:
@OrenMe @dan-ziv agreed & changed

src/plugin/plugin-manager.js
```diff
@@ -82,12 +82,12 @@ export default class PluginManager {
...
     }
     let pluginClass = PluginManager._registry.get(name);
-    if (pluginClass && pluginClass.isValid()) {
+    if (pluginClass && pluginClass.isValid() && !config.disable) {
       this._plugins.set(name, pluginClass.createPlugin(name, player, config));
       logger.debug(`Plugin <${name}> has been loaded`);
       return true;
     }
-    logger.debug(`Plugin <${name}> isn\'t loaded, isValid()=false`);
+    logger.debug(`Plugin <${name}> isn\'t loaded, isValid()=false or it's disabled`);
```
OrenMe: `isValid()=false, disabled=${config.disable}` will make it less guessing work when reading a log

dan-ziv:
@odedhutzler need to update docs and explain regarding this option

OrenMe:
@dan-ziv docs added in #233 please review and release request for changes

PR #230, opened 2018-05-27
fix(FEC-8255): video starts from beginning instead 12th sec

dan-ziv:

set the current time just after loadeddata event is raised.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -211,7 +211,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
   load(startTime: ?number): Promise<Object> {
     if (!this._loadPromise) {
       this._loadPromise = new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -220,10 +220,6 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           this._videoElement.src = this._sourceObj.url;
           this._trigger(CustomEventType.ABR_MODE_CHANGED, {mode: this._isProgressivePlayback() ? 'manual' : 'auto'});
         }
-        if (startTime && startTime > -1) {
-          this._videoElement.currentTime = startTime;
-        }
-        this._videoElement.load();
```
OrenMe: @dan-ziv how do you account for removing call to `load`? what was the use case for previously having it here?

OrenMe: @dan-ziv - so we don't need it here?

dan-ziv: I don't think so, but lets keep it.
Plz review

PR #234, opened 2018-05-30
fix(FEC-8272): sometimes when seeking spinner display constantly in IE11

odedhutzler:

Add time update listener to the buffering state

when buffering, if we get a timeupdate event, we need to set the state of the player to playing. (fixes a bug in IE11) the bug is that there is no playing event sent sometimes after it's ended

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/state/state-manager.js
+++ b/src/state/state-manager.js
@@ -48,7 +48,7 @@ export default class StateManager {
    */
   _prevState: State | null;
   /**
-   * Holds the time of the last seek
+   * Holds the time of the beginning of the last buffering (waiting event)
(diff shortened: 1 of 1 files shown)


src/state/state-manager.js
```diff
@@ -47,6 +47,13 @@ export default class StateManager {
    * @private
    */
   _prevState: State | null;
+  /**
+   * Holds the time of the last seek
```
yairans: fix jsdoc

OrenMe:
@yairans please re-review

PR #239, opened 2018-06-14
feat(FEC-8046): 360 support

yairans:

* expose VR API's and event
* enable to pass options to event target (needed to pass `passive` by 'vr' plugin) 
* enrich the Dom util

### CheckLists

- [ ] Docs have been updated


src/player.js
```diff
@@ -1205,6 +1205,27 @@ export default class Player extends FakeEventTarget {
...
+  // </editor-fold>
+
+  // <editor-fold desc="Plugins API">
+
+  /**
+   * get a plugin instance
+   * @param {string} name - the plugin name
+   * @returns {?BasePlugin} - the plugin instance
+   */
+  getPlugin(name: string): ?BasePlugin {
```
dan-ziv: If we are already adding this API we need to add also something like
```js
isPluginActive(name) : boolean
```

src/error/category.js
```diff
@@ -28,7 +28,10 @@ const Category: CategoryType = {
   'ADS': 8,
 
   /** Errors in the database storage (offline). */
-  'STORAGE': 9
+  'STORAGE': 9,
+
+  /** Errors from a plugin. */
+  'PLUGIN': 11
```
OrenMe: no need for plugin name-spacing, too generic to have the  `PLUGIN_FATAL_ERROR`, we already have a fatal error convention via the `SEVERITY` so it's also redundant to create an event with `FATAL` naming IMO.

src/player.js
OrenMe: I see a case of exposing the plugins outside, think of cases where we are also using a 3rd party API that might be needed for application integration(hls.js needed by a 3rd party we don't implement, IMA is needed for MOAT integration etc.)

As we are adding VR specific stuff I would like to keep it focused and not add unplanned APIs.
@yairans add the `toggleVrStereoMode` here and refer it to the plugin directly - via the plugin manager, and change the UI to call `player.toggleVrStereoMode()`.
You can also remove the kaltura player bindings then.
Once we create the controller for VR we will change this implementation to use it.

OrenMe: also, maybe toggle is not the best option, as it forces you to know the state, so you will have to also expose `isStereoModeSet`

PR #249, opened 2018-07-01
fix: adding external captions default ability

odedhutzler:

checking the 'default' attribute when adding a caption.

updated the docs / type.

### CheckLists

- [ ] Docs have been updated


src/track/external-captions-handler.js
```diff
@@ -126,7 +126,7 @@ class ExternalCaptionsHandler extends FakeEventTarget {
     const newTextTracks = [];
     captions.forEach(caption => {
       const track = new TextTrack({
-        active: false,
+        active: caption.default ? true : false,
```
OrenMe: the expression should be boolean so can be changed to `!!caption.default`

PR #259, opened 2018-07-04
fix(FEC-8376): the menu ui doesn't show the right caption after selection

odedhutzler:
updating the indexes only after there are tracks in the player.
updating the indexes only after there are tracks in the player and also notifying the ui on the indexes change.

Please add a detailed description of the change, weather it's an enhancement or a bugfix.

### CheckLists

- [ ] Docs have been updated


src/player.js
```diff
@@ -1517,7 +1517,10 @@ export default class Player extends FakeEventTarget {
           const videoElement = this.getVideoElement();
           return videoElement ? Array.from(videoElement.textTracks).findIndex(track => track ? track.language === textTrack.language : false) : -1;
         };
-        this._getTracksByType(TrackType.TEXT).forEach(track => track.index = getNativeLanguageTrackIndex(track));
+        this.ready().then(() => {
```
OrenMe: If we already need to wait for ready event to map the index than it probably means that we don’t need to be handled here at all and move this to the ready promise chain just before we emit the regular TEXT_TRACK_CHANGED

odedhutzler: @OrenMe might there be a case when the ready promise will resolved before we add native tracks?

OrenMe: External native tracks are controlled by you and not by 3rd party lib that parses and downloads a manifest so this is something you have full control over. AFAICT no need for guessing work here

OrenMe: @odedhutzler what is the status of this?

PR #261, opened 2018-07-08
fix(FEC-8386): Kava analytics retrieved by default on OTT

yairans:

Keep the plugin's `disable` value for the next `load` calls

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/plugin/plugin-manager.js
+++ b/src/plugin/plugin-manager.js
@@ -98,7 +98,7 @@ export default class PluginManager {
       logger.debug(`Plugin <${name}> has been loaded`);
       return true;
     }
-    logger.debug(`Plugin <${name}> isn\'t loaded, isValid()=${isValidPlugin ? 'true' : 'false'}, disabled=${isDisablePlugin ? 'true' : 'false'}`);
+    logger.debug(`Plugin <${name}> isn\'t loaded, isValid()=${isValidPlugin.toString()}, disabled=${isDisablePlugin.toString()}`);
     return false;
   }
 
--- a/src/plugin/plugin-manager.js
+++ b/src/plugin/plugin-manager.js
@@ -32,6 +32,8 @@ export default class PluginManager {
   _plugins: Map<string, BasePlugin> = new Map();
   /**
    * Is disabled plugin map.
+   * Maps plugin's name to a boolean.
+   * false means the plugin is disable. true or plugin name doesn't exist in the map means the plugin is not disable.
    * @type {Map}
    * @private
    */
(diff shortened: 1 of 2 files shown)


src/plugin/plugin-manager.js
```diff
@@ -82,12 +88,17 @@ export default class PluginManager {
       throw new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.RUNTIME_ERROR_NOT_REGISTERED_PLUGIN, name);
     }
     let pluginClass = PluginManager._registry.get(name);
-    if (pluginClass && pluginClass.isValid() && !config.disable) {
+    if (typeof config.disable === "boolean") {
+      this._isDisabledPluginMap.set(name, config.disable);
+    }
+    const isDisablePlugin = !!this._isDisabledPluginMap.get(name);
```
dan-ziv: Use the `has` API for map and you won't need the casting to boolean operation.

src/plugin/plugin-manager.js
```diff
@@ -82,12 +88,17 @@ export default class PluginManager {
...
+    const isValidPlugin = pluginClass ? pluginClass.isValid() : false;
+    if (pluginClass && isValidPlugin && !isDisablePlugin) {
       this._plugins.set(name, pluginClass.createPlugin(name, player, config));
       logger.debug(`Plugin <${name}> has been loaded`);
       return true;
     }
-    logger.debug(`Plugin <${name}> isn\'t loaded, isValid()=false, disabled=${config.disable}`);
+    logger.debug(`Plugin <${name}> isn\'t loaded, isValid()=${isValidPlugin ? 'true' : 'false'}, disabled=${isDisablePlugin ? 'true' : 'false'}`);
```
dan-ziv: `isValidPlugin.toString()`
`isDisablePlugin.toString()`

src/plugin/plugin-manager.js
```diff
@@ -30,6 +30,12 @@ export default class PluginManager {
    * @private
    */
   _plugins: Map<string, BasePlugin> = new Map();
+  /**
+   * Is disabled plugin map.
+   * @type {Map}
+   * @private
+   */
+  _isDisabledPluginMap: Map<string, boolean> = new Map();
```
OrenMe: `disabledPlugins` Is better fitting and u can save the bytes and set it as an array as the array only needs to have disabled plugins in it so you can use `disabledPlugins.contans(pluginNmae)`

OrenMe: Consider that this implementation is a bitconfusing cause if a plugin was not set with disable as Boolean  it will not be in the map at all and this is a bit misleading.
If it is not in it what does it mean? If it is and has true what does it mean. Leads to guessing.
As I already mentioned in the closed PR in Kaltura player we will add an API to enable/disable and get state of of plugins from plugins manager, so this is a good time to think about it.
If you don’t have a better idea at the moment at least make sure to add a plugin if it was init to the map with value false so we have consistency.

PR #264, opened 2018-07-11
feat(FEC-8394): move rates to engine

itaykinnrot:
remove the hard coded rates and move them to the player engine + support different rates for live and vod (remove this logic from the ui...)

### CheckLists

- [X] Travis tests are passing (or test results are not worse than on master branch :))
- [ ] Docs have been updated


src/player.js
```diff
@@ -42,19 +42,7 @@ import PKError from './error/error'
...
- * @type {Array<number>}
- * @const
- */
-const PLAYBACK_RATES = [0.5, 1, 2, 4];
 
-/**
- * The player default playback rate.
- * @type {number}
- * @const
- */
```
odedhutzler: why delete that const?

src/engines/html5/html5.js
```diff
@@ -149,6 +149,20 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+   * The player playback rates.
+   * @type {Array<number>}
+   */
+   PLAYBACK_RATES = [0.5, 1, 2, 4];
+
+   /**
+   * The player playback rates.
+   * @type {Array<number>}
+   */
+   LIVE_PLAYBACK_RATES = [1];
```
yairans: should be `static const`

yairans: also fix the jsdoc. also above

src/engines/html5/html5.js
```diff
@@ -840,6 +854,14 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     return this._el.getAttribute('crossorigin');
   }
 
+  /**
+   * get the playback rates
+   * @return {number[]} - plyback rates
```
yairans: `Array<number>`. typo - `plyback `

src/engines/html5/html5.js
```diff
@@ -840,6 +854,14 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+   */
+  get playbackRates(): Array<number> {
+    return this.isLive()? this.LIVE_PLAYBACK_RATES : this.PLAYBACK_RATES;
```
yairans: should be static.

src/engines/html5/html5.js
OrenMe: @itaykinnrot why differentiate live and vod?

src/engines/html5/html5.js
OrenMe: again, why differentiate?

src/engines/html5/html5.js
itaykinnrot: @OrenMe  to keep the same behaviour we have today - i removed it from the ui

OrenMe: The reason it was in the UI is that we don’t want to limit the API of player with BL.
The UI in this case is opinionated and can make assumptions but from playback perspective having rate higher or lower than 1 for live is also possible(DVR playback).
Please keep this logic in UI.

PR #265, opened 2018-07-12
feat(FEC-8390, FEC-8246): support 608/708 captions

yairans:
* add config for CEA captions
* allow text track without language or label. but filter metadata.
* call to `_maybeSetTracksLabels` and `_setDefaultTracks` in any `TRACKS_CHANGED` event

depends on https://github.com/kaltura/playkit-js-ui/pull/254 

### CheckLists

- [ ] Docs have been updated


src/assets/style.css
```diff
@@ -44,7 +44,7 @@
 
 .playkit-subtitles {
   position: absolute;
-  top: 0;
+  top: 60px;
```
OrenMe: @yairans what about captions that want to be positioned in upper part of div? now they will always have -60px offset from top

src/player.js
```diff
@@ -1849,8 +1847,10 @@ export default class Player extends FakeEventTarget {
    */
   _updateTracks(tracks: Array<Track>): void {
     Player._logger.debug('Tracks changed', tracks);
-    this._tracks = [...tracks, ...this._externalCaptionsHandler.getExternalTracks(tracks)];
+    this._tracks = tracks.concat(this._externalCaptionsHandler.getExternalTracks(tracks));
```
OrenMe: what is the difference?

src/player.js
```diff
@@ -1849,8 +1847,10 @@ export default class Player extends FakeEventTarget {
...
     this._addTextTrackOffOption();
+    this._maybeSetTracksLabels();
```
OrenMe: why move this inside the method and not keep it as before? because 608/708 can update on the fly and you want to call `_updateTracks` and apply `_maybeSetTracksLabels` and `_setDefaultTracks`.
You need to check the `_setDefaultTracks` cause you only want it to happen once and in this case if there's some unwanted side effect it might be called several times, when ever `_onTracksChanged` is called.

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -415,7 +415,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           language: textTracks[i].language,
           index: i
         };
-        if (settings.language || settings.label) {
```
OrenMe: what happens in this case? if 608/708 caption doesn't have a language/label what will be shown?
Do we need a setting option like hls.js expose?

OrenMe: also, the test for this now fails

src/assets/style.css
yairans: right. i don't have a better solution

src/player.js
yairans: I've checked this change of course.
it behaves correct thanks `this._playbackAttributesState.textLanguage`

src/player.js
yairans: Do you really prefer the previous??

OrenMe: No preference, but the previous commit made this change, and now you are reverting it.
If this is just opinionated change please consult with whomever made the original change or at least explain why you reverted the code.
Is this pref related?

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -415,7 +422,12 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+        if (settings.kind === 'subtitles') {
+          parsedTracks.push(new PKTextTrack(settings));
+        }
+        if (settings.kind === 'captions' && this._config.enableCEA708Captions) {
+          settings.label = settings.label || captionsTextTrackLabels.pop();
```
OrenMe: can we do better then pop-ing the array? seems a bit hack-ish.

src/assets/style.css
OrenMe: what happens in native implementations? what does safari do?

src/player.js
yairans: The original was only `this._tracks = tracks`. while the external captions change it got to the red line above, and i'm changing the green line above. same logic but readable.

src/engines/html5/media-source/adapters/native-adapter.js
yairans: what bad with `pop` (actually should be `shift`)? it'a a legitimate api.

src/assets/style.css
yairans: we have a problem in native impl....

OrenMe: Can u elaborate? Also, what happens with native player of safari when stream has captions in upper part of video?

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -415,7 +424,12 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           language: textTracks[i].language,
           index: i
         };
-        if (settings.language || settings.label) {
+        if (settings.kind === 'subtitles') {
+          parsedTracks.push(new PKTextTrack(settings));
+        }
+        if (settings.kind === 'captions' && this._config.enableCEA708Captions) {
```
dan-ziv: else if

PR #279, opened 2018-08-15
fix(FEC-8457): autoplay check fails due to browser CPU/load issue

odedhutzler:
Using a timeout to determine we cannot autoplay cause issues in IE11 (shows we cannot, when we can). Reference for the code change: https://github.com/video-dev/can-autoplay#videooptions

The changes:
- Using a blob to test autoplay.
- When the element.play() does not return a promise check:
1. if it's paused and return that autoplay is not available. (old iOS)
2. if it's not - then return true (IE 11)

### CheckLists

- [ ] Docs have been updated


src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,16 +1,30 @@
...
..., 0, 0, 1, 0, 0, 0, 12, 117, 114, 108, 32, 0, 0, 0, 1, 0, 0, 0, 211, 115, 116, 98, 108, 0, 0, 0, 103, 115, 116, 115, 100, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 87, 109, 112, 52, 97, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 16, 0, 0, 0, 0, 172, 68, 0, 0, 0, 0, 0, 51, 101, 115, 100, 115, 0, 0, 0, 0, 3, 128, 128, 128, 34, 0, 2, 0, 4, 128, 128, 128, 20, 64, 21, 0, 0, 0, 0, 1, 244, 0, 0, 1, 243, 249, 5, 128, 128, 128, 2, 18, 16, 6, 128, 128, 128, 1, 2, 0, 0, 0, 24, 115, 116, 116, 115, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 4, 0, 0, 0, 0, 28, 115, 116, 115, 99, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 1, 0, 0, 0, 28, 115, 116, 115, 122, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 1, 115, 0, 0, 1, 116, 0, 0, 0, 20, 115, 116, 99, 111, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 44, 0, 0, 0, 98, 117, 100, 116, 97, 0, 0, 0, 90, 109, 101, 116, 97, 0, 0, 0, 0, 0, 0, 0, 33, 104, 100, 108, 114, 0, 0, 0, 0, 0, 0, 0, 0, 109, 100, 105, 114, 97, 112, 112, 108, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 45, 105, 108, 115, 116, 0, 0, 0, 37, 169, 116, 111, 111, 0, 0, 0, 29, 100, 97, 116, 97, 0, 0, 0, 1, 0, 0, 0, 0, 76, 97, 118, 102, 53, 54, 46, 52, 48, 46, 49, 48, 49])], {type: 'video/mp4'});
```
yairans: move to file

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -1,16 +1,30 @@
 // @flow
 import * as Utils from '../../../utils/util';
-import * as EncodingSources from '../../../assets/encoding-sources.json';
 import {Html5EventType} from '../../../event/event-type';
```
yairans: delete `encoding-sources.json` file if unused

odedhutzler: fixed

src/engines/html5/capabilities/html5-autoplay.js
odedhutzler: fixed

src/engines/html5/capabilities/html5-autoplay.js
```diff
@@ -56,6 +64,9 @@ export default class Html5AutoPlayCapability implements ICapability {
    * @private
    */
   static _getPlayPromise(): Promise<*> {
+    testVideoElement.addEventListener(Html5EventType.ERROR, () => {
+      return Promise.reject();
```
OrenMe: This promise reject is not part of the `_forcePromiseReturnValue` promise you create there so 1. it won't solve the promise you need and 2. as you return this from an event listener this will just emit a global promise rejection not tied to anything.
Need to listen to error event in the `_forcePromiseReturnValue` and use the reject you got from the promise you created there.
Some browsers have different behaviours, not all will have error events I think so need to also listen to pause and abort possibly(I think FF had this issue)

PR #280, opened 2018-08-15
fix(FEC-8143): The video starts with 'Eng' captions although configuration set to 'Rus' on Native

odedhutzler:
we set the `track.mode` (of `element.textTracks`) to `hidden` or `showing`, when it's active.  
so we need to check also `hidden` (cause it also means the `textTrack` is active.)

### CheckLists

- [ ] Docs have been updated


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -640,7 +640,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     const getActiveVidTextTrackIndex = () => {
       for (let i = 0; i < this._videoElement.textTracks.length; i++) {
         const textTrack = this._videoElement.textTracks[i];
-        if (textTrack.mode === 'showing') {
+        if (['showing', 'hidden'].includes(textTrack.mode)) {
```
OrenMe: Better abstract the code that decides what string is used for display mode:
```javascript
_getDisplayTextTrackModeString () {
  return this._config.displayTextTrack ? 'showing' : 'hidden';
}
```

and use it here and in the other place we use it in the code: https://github.com/kaltura/playkit-js/blob/3503b94373902728465078776af7b02d68fb5aa5/src/engines/html5/media-source/adapters/native-adapter.js#L598

PR #283, opened 2018-08-16
fix(FEC-7594): native adapter keeps buffering when connection is lost

odedhutzler:
When there is no network and native adapter is used, the video element keeps buffering forever (the video element state stays '`waiting`'). 

the problem is that no error is thrown, so the player doesn't know there is an issue and showing the spinner forever.

the suggested fix starts a timeout upon a '`waiting`' event and checking if the browser is offline.
the timeout is cleared if we are playing / pausing etc.

### CheckLists

- [ ] Docs have been updated


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -283,6 +289,33 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+          )
+        );
+      } else {
+        this._bufferTimeoutId = setTimeout(onTimeOut, WAIT_TIME);
+      }
+    };
+    this._bufferTimeoutId = setTimeout(onTimeOut, WAIT_TIME);
+  }
+
+  _clearBufferTimeout(): void {
```
yairans: consider `_clearWaitingTimeout`

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -283,6 +289,33 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+    if (this._bufferTimeoutId) {
+      clearTimeout(this._bufferTimeoutId);
```
yairans: `this._bufferTimeoutId = null`

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -283,6 +289,33 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+          Html5EventType.ERROR,
+          new Error(
+            Error.Severity.CRITICAL,
+            Error.Category.NETWORK,
+            Error.Code.TIMEOUT,
+            `player exceeded ${WAIT_TIME} timeout and connection seems to be offline`
+          )
+        );
+      } else {
+        this._bufferTimeoutId = setTimeout(onTimeOut, WAIT_TIME);
```
yairans: use `setInterval` instead of nested `setTimeout`

OrenMe:
@yairans please update review pending @odedhutzler's changes.

PR #288, opened 2018-09-02
feat: ads controller

yairans:
* expose a new api `ads` for the ads manipulation (`player.ads.skipAd()` etc.)
* expose new events `PLAYBACK_START` and `PLAYBACK_ENDED` (include ads)
* change `PLAYBACK_STARTED` to `FIRST_PLAYING`

### CheckLists

- [ ] Docs have been updated


src/event/event-type.js
```diff
@@ -180,6 +180,10 @@ const CustomEventType: EventTypes = {
    * Fires on the first time that playback has been started.
    */
   PLAYBACK_STARTED: 'playbackstarted',
+  /**
+   * Fires when the playback (includes postrolls) is ended.
+   */
+  PLAYBACK_ENDED: 'playbackended',
```
yairans: Sounds as the Inverse of `PLAYBACK_STARTED` but it's not...

src/player.js
```diff
@@ -1715,6 +1733,15 @@ export default class Player extends FakeEventTarget {
     };
   }
 
+  _maybeCreateAdsController(): void {
+    if (!this._adsController) {
+      const adsPluginController = this._controllerProvider.getAdsController();
+      if (adsPluginController) {
+        this._adsController = new AdsController(this, adsPluginController);
```
OrenMe: doesn't seem to use both `player` and `adsPluginController`

src/ads/ads-controller.js
```diff
@@ -0,0 +1,68 @@
...
+  _adsPluginController: IAdsController;
+  _allAdsCompleted: boolean;
+  _eventManager: EventManager;
+
+  /**
+   * The ads controller.
+   * @param {Player} player - The player.
+   * @param {IAdsController} adsPluginController - the controller of the current plugin instance.
+   */
+  constructor(player: Player, adsPluginController: IAdsController) {
```
OrenMe: doesn't seem to use both `player` and `adsPluginController`

src/event/event-type.js
OrenMe: `PLAYBACK_STARTED` is ok where it is.
It means that actual video frame is being rendered on screen.
Now that we have the ads playback taken into account it just means that we need to listen to ad_playing and fire it as well.
We already have a boolean flag to fire it only on first time so just extract this to a method, that is called from playing event and ad_playing event - first one reaching it will emit it and set the boolean flag to true.

src/ads/ads-controller.js
yairans: `player` used for the events. `adsPluginController` used for ads API e.g `adsPluginController.skipAd()`

src/player.js
yairans: see my comment below

flow-typed/interfaces/ads-controller-provider.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+declare interface IAdsControllerProvider {
```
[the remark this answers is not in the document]
yairans: no

src/controller/controller-provider.js
```diff
@@ -0,0 +1,31 @@
+//@flow
+
+import PluginManager from '../plugin/plugin-manager';
+
+/**
+ * Controller provider
+ * @classdesc
+ */
+class ControllerProvider {
```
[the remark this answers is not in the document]
yairans: There is no `IControllerProvider` interface

PR #293, opened 2018-10-07
fix(FEC-7594): native adapter does not raise an error in case network disconnects.

odedhutzler:

When network disconnects, IE triggers `stall` event, and after the buffer ends it keeps trigger `timeupdate` events (until the network comes back online.. or until forever...) this fix checks if the current time update is greater than the last one - so we know we aren't stuck, and raise an Error.

IE also doesn't change it's state to paused / buffering so we also fire `waiting` event if we are still on playing state but the current time of the player does not change. We trigger `playing` once the player resume playing

### CheckLists

- [ ] Docs have been updated


odedhutzler:
@OrenMe @dan-ziv @yairans 

not sure which event should the native adapter raise in such case - buffering? so the player will change it state.

PR #295, opened 2018-10-17
feat(FEC-7808): quality watcher for the player

odedhutzler:

The engine will initialize the `dropped-frames-watcher` class that will start an interval to check dropped frames on the video via the adapter (added an API in the adapter as well - `getCurrentQuality`.
If there are too many frame drops it will raise an `EXCEEDED_MAX_FRAME_DROP` event that will bubble to to the player which will change to a lower bandwidth video track (if it can).

### CheckLists

- [ ] Docs have been updated


src/player.js
```diff
@@ -1508,6 +1503,14 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  _handleFrameDrop(event: FakeEvent): void {
+    const sortedVideoTracks = this._getTracksByType(TrackType.VIDEO).sort((trackA, trackB) => trackA.bandwidth > trackB.bandwidth);
+    const currentBandwidthTrackIndex = sortedVideoTracks.findIndex(track => track.bandwidth === event.payload.bandwidth);
+    if (currentBandwidthTrackIndex - 1 >= 0) {
```
dan-ziv: findIndex returns -1 in case of no matching track.
`currentBandwidthTrackIndex !== -1` will be more readable condition.

src/event/event-type.js
```diff
@@ -201,7 +201,11 @@ const CustomEventType: PKEventTypes = {
   /**
    * Fired when the vr stereo mode changed
    */
-  VR_STEREO_MODE_CHANGED: 'vrstereomodechanged'
+  VR_STEREO_MODE_CHANGED: 'vrstereomodechanged',
+  /**
+   * Fired when too many frames are dropped
+   */
+  EXCEEDED_MAX_FRAME_DROP: 'exceededmaxframedrop'
```
dan-ziv: This is an internal event and not a public player event. you shouldn't expose it here but do define it in the quality watcher class.

src/engines/html5/html5.js
```diff
@@ -78,6 +79,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @static
    */
   static _el: HTMLVideoElement;
+  _droppedFramesWatcher: ?droppedFramesWatcher = null;
```
dan-ziv: 1. Define it upper the static members with all others instance members.
2. Why define it as nullable if you don't do it anywhere?

src/engines/dropped-frames-watcher.js
```diff
@@ -0,0 +1,75 @@
+// @flow
+import FakeEventTarget from '../event/fake-event-target';
+import FakeEvent from '../event/fake-event';
+import {CustomEventType} from '../event/event-type';
+import getLogger from '../utils/logger';
+import MediaSourceProvider from './html5/media-source/media-source-provider';
+
+export default class droppedFramesWatcher extends FakeEventTarget {
```
dan-ziv: Class should start with upper case -> `DroppedFramesWatcher `

dan-ziv: Don't export default but named export -> 
`class DroppedFramesWatcher extends FakeEventTarget { ... }`
`export {DroppedFramesWatcher};`

src/engines/dropped-frames-watcher.js
```diff
@@ -0,0 +1,74 @@
...
+  _config: Object = {};
+  _videoElement: HTMLVideoElement;
+  static _logger: any = getLogger('droppedFramesWatcher');
+
+  constructor(mediaSourceAdapter: IMediaSourceAdapter, config: Object, videoElement: HTMLVideoElement) {
+    super();
+    this._mediaSourceAdapter = mediaSourceAdapter;
+    this._config = config;
+    this._videoElement = videoElement;
+    this._droppedFramesInterval = setInterval(() => this._maybeAdjustBitrateQuality(), this._config.fpsDroppedFramesInterval);
```
OrenMe: `_maybeAdjustBitrateQuality` is not aproper name as it doesn't adjust bitrate(?~) but just checks for it

src/player.js
```diff
@@ -1508,6 +1504,16 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  _handleFrameDrop(event: FakeEvent): void {
+    const videoTracks = ((this._getTracksByType(TrackType.VIDEO): any): Array<VideoTrack>);
+    const sortedVideoTracks = videoTracks.sort((trackA, trackB) => trackA.bandwidth - trackB.bandwidth);
+    const currentBandwidthTrackIndex = sortedVideoTracks.findIndex(track => track.bandwidth && track.bandwidth === event.payload.bandwidth);
+    // not out of the array boundaries
+    if (currentBandwidthTrackIndex - 1 >= 0) {
+      this.selectTrack(sortedVideoTracks[currentBandwidthTrackIndex - 1]);
```
OrenMe: This doesn't enable adaptive switching, but just select a specific video track, which wasn't the intended use case.
The requirement is for adaptivness to still be retained while marking a level as not allowed for use due to high FPS.

src/engines/dropped-frames-watcher.js
```diff
@@ -0,0 +1,74 @@
...
+      if (decodedFrames) {
+        if (this._lastTime) {
+          const currentPeriod = currentTime - this._lastTime,
+            currentDropped = droppedFrames - this._lastDroppedFrames,
+            currentDecoded = decodedFrames - this._lastDecodedFrames,
+            droppedFPS = (1000 * currentDropped) / currentPeriod;
+          if (droppedFPS > 0) {
+            DroppedFramesWatcher._logger.debug('checkFPS : droppedFPS/decodedFPS:' + droppedFPS / ((1000 * currentDecoded) / currentPeriod));
+            if (currentDropped > this._config.fpsDroppedMonitoringThreshold * currentDecoded) {
+              const currentBandwidth = this._mediaSourceAdapter.getCurrentQuality();
```
OrenMe: I do not see a reason to use this API like this as the data is already saved in player.
You can use an event and pass the player instance to listen to it on track change.
Also exposing this like this is very limited anyway - as if you have an API called `getCurrentQuality` you would expect to have the entire level object and be able to access level.bitrate on it.

src/engines/dropped-frames-watcher.js
```diff
@@ -0,0 +1,69 @@
...
+      if (decodedFrames) {
+        if (this._lastTime) {
+          const currentPeriod = currentTime - this._lastTime,
+            currentDropped = droppedFrames - this._lastDroppedFrames,
+            currentDecoded = decodedFrames - this._lastDecodedFrames,
+            droppedFPS = (1000 * currentDropped) / currentPeriod;
+          if (droppedFPS > 0) {
+            DroppedFramesWatcher._logger.debug('checkFPS : droppedFPS/decodedFPS:' + droppedFPS / ((1000 * currentDecoded) / currentPeriod));
+            if (currentDropped > this._config.fpsDroppedMonitoringThreshold * currentDecoded) {
+              this._mediaSourceAdapter.restrictBandwidth();
```
OrenMe: `restrictBandwidth` is very coupled to specific functionality of what you are trying to achieve in this issue, but in general what you want to have is ABR control API - so `setMaxBitrate` is more suitable and more generic.
You will however need to save a reference to current track which you can get via the event handler fro the adapter

src/engines/html5/media-source/base-media-source-adapter.js
```diff
@@ -39,6 +39,12 @@ export default class BaseMediaSourceAdapter extends FakeEventTarget implements I
    */
   _videoElement: HTMLVideoElement;
 
+  /**
+   * The adapter capabilites
+   * @private
+   */
+  _capabilites: Object = {fpsControl: false};
```
dan-ziv: Define a type for this.
For example:
```MediaSourceCapabilities```

File: src/engines/html5/html5.js [hunk 10]
```diff
@@ -823,6 +827,9 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   _init(source: PKMediaSourceObject, config: Object): void {
     this._config = config;
     this._loadMediaSourceAdapter(source);
+    if (this._mediaSourceAdapter) {
```
dan-ziv: Move this line to be inside ```_loadMediaSourceAdapter``` since you always need it incase MSE is exists.
Then you won't need the condition.

src/engines/dropped-frames-watcher.js
```diff
@@ -0,0 +1,94 @@
...
+        this._lastDroppedFrames = droppedFrames;
+        this._lastDecodedFrames = decodedFrames;
+      }
+    } catch (error) {
+      DroppedFramesWatcher._logger.error('Error occur while trying to check dropFrames: ', error);
+    }
+  }
+
+  destroy(): void {
+    if (this._droppedFramesInterval) {
```
OrenMe: destroy event manager

PR #297, opened 2018-10-24
fix(FEC-8554): update captions on resize event

odedhutzler:

Added resize listener to the window, and updated the captions accordingly.
The `_resetTextCuesAndReposition` function acts as a `debounce` function, which will update the captions only after the last resize.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/test/src/player.spec.js
+++ b/test/src/player.spec.js
@@ -2857,6 +2857,32 @@ describe('Player', function() {
     });
   });
 
+  describe('on resize event', function() {
+    let config, player, sandbox;
+
+    beforeEach(() => {
+      sandbox = sinon.sandbox.create();
+      config = getConfigStructure();
+      config.sources = sourcesConfig.Mp4;
+      player = new Player(config);
+    });
+
+    afterEach(() => {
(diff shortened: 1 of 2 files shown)


odedhutzler:
@OrenMe no need to listen for `orientationchange`

odedhutzler:
added unit test

test/src/player.spec.js
```diff
@@ -2857,6 +2857,32 @@ describe('Player', function() {
...
+      config.sources = sourcesConfig.Mp4;
+      player = new Player(config);
+    });
+
+    afterEach(() => {
+      sandbox.restore();
+      player.destroy();
+    });
+
+    it('should call _resetTextCuesAndReposition function', done => {
```
OrenMe: In general I think the correct title should be more declarative of what is happening, e.g. - 

> should re-calculate size and position of text cue

odedhutzler: but it just calls that function and that's what the unit test checks.

there is another unit test that checks the _resetTextCuesAndReposition function itself.

PR #298, opened 2018-10-28
feat(FEC-8277): Picture in Picture support

odedhutzler:

Adding Picture in picture API:
```
enterPictureInPicture()
exitPictureInPicture()
isInPictureInPicture()
isPictureInPictureSupported()
```

Exposing 3 events (HTML5 events):
Chrome events:
`enterpictureinpicture` (after entering PIP)
`leavepictureinpicutre` (after exit PIP)
Safari event:
`webkitpresentationmodechanged` (when transitioning from inline to PIP / PIP to inline)
### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -53,7 +53,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @type {boolean}
    * @private
    */
-  _isInPip: boolean = false;
+  _isInPictureInPicture: boolean = false;
 
   /**
    * The html5 class logger.
@@ -948,7 +948,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     return extended;
   }
 
-  enterPip(): void {
+  enterPictureInPicture(): void {
     try {
       // Currently it's supported in chrome and in safari. So if we consider checking support before,
       // we can use this flag to distinguish between two. In the future we might need a different method.
@@ -961,13 +961,13 @@ export default class Html5 extends FakeEventTarget implements IEngine {
       } else {
         this._el.webkitSetPresentationMode('picture-in-picture');
       }
-      this._isInPip = true;
+      this._isInPictureInPicture = true;
     } catch (error) {
       this.dispatchEvent(new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.CANNOT_ENTER_PICTURE_IN_PICTURE, error));
     }
   }
 
-  exitPip(): void {
+  exitPictureInPicture(): void {
     try {
       // Currently it's supported in chrome and in safari. So if we consider checking support before,
       // we can use this flag to distinguish between two. In the future we might need a different method.
@@ -980,21 +980,24 @@ export default class Html5 extends FakeEventTarget implements IEngine {
       } else {
         this._el.webkitSetPresentationMode('inline');
       }
-      this._isInPip = false;
+      this._isInPictureInPicture = false;
     } catch (error) {
       this.dispatchEvent(new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.CANNOT_ENTER_PICTURE_IN_PICTURE, error));
     }
   }
 
-  isPipSupported(): boolean {
-    if (document.pictureInPictureEnabled || (this._el.webkitSupportsPresentationMode && typeof this._el.webkitSetPresentationMode === 'function')) {
+  isPictureInPictureSupported(): boolean {
+    if (
+      document.pictureInPictureEnabled ||
+      (this._el.webkitSupportsPresentationMode && this._el.webkitSupportsPresentationMode('picture-in-picture'))
+    ) {
       return true;
     } else {
       return false;
     }
   }
 
-  get isInPip(): boolean {
-    return this._isInPip;
+  get isInPictureInPicture(): boolean {
+    return this._isInPictureInPicture;
   }
 }
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -252,6 +252,8 @@ export default class Html5 extends FakeEventTarget implements IEngine {
         }
       });
     });
+    this._eventManager.listen(this._el, Html5EventType.ENTER_PICTURE_IN_PICTURE, () => (this._isInPictureInPicture = true));
+    this._eventManager.listen(this._el, Html5EventType.LEAVE_PICTURE_IN_PICTURE, () => (this._isInPictureInPicture = false));
     if (this._mediaSourceAdapter) {
       this._eventManager.listen(this._mediaSourceAdapter, CustomEventType.VIDEO_TRACK_CHANGED, (event: FakeEvent) => this.dispatchEvent(event));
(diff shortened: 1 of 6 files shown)


src/engines/html5/html5.js
```diff
@@ -941,4 +941,50 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     }
     return extended;
   }
+
+  enterPip(): void {
```
dan-ziv: Change to ```enterPictureInPicture```
Pip is not a known acronyms, but if you already use it so use capital letters -> PIP

src/player.js
```diff
@@ -1210,6 +1216,55 @@ export default class Player extends FakeEventTarget {
...
+  /**
+   * Check if picture in picture supported in this environment
+   * @public
+   * @return {boolean} if the picture in picture feature is supported in this environment
+   */
+  isPictureInPictureSupported(): boolean {
+    return this._engine.isPipSupported();
+  }
+
+  _handlePipEvent(event: FakeEvent): void {
```
dan-ziv: Move down to private methods!

src/player.js
```diff
@@ -1210,6 +1216,55 @@ export default class Player extends FakeEventTarget {
...
+    this._isInPip = !this._isInPip;
```
dan-ziv: Don't change the state on the event itself, this is very dangerous.
If someone else will register to this event before this listener it will get a wrong state.
Update the state in the method which triggers the event.

src/engines/html5/html5.js
odedhutzler: changed

src/player.js
odedhutzler: not relevant (method removed)

src/player.js
odedhutzler: not relevant (moved to the engine)

src/engines/html5/html5.js
```diff
@@ -941,4 +947,57 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
     return extended;
   }
+
+  enterPictureInPicture(): void {
+    try {
+      // Currently it's supported in chrome and in safari. So if we consider checking support before,
+      // we can use this flag to distinguish between two. In the future we might need a different method.
+      if (document.pictureInPictureEnabled) {
+        this._el
+          .requestPictureInPicture()
```
OrenMe: use this promise and add `then` to set `_isInPictureInPicture` to true and false in `catch` if it throws

src/engines/html5/html5.js
```diff
@@ -941,4 +947,57 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+          .catch(error =>
+            this.dispatchEvent(new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.CANNOT_ENTER_PICTURE_IN_PICTURE, error))
+          );
+      } else {
+        this._el.webkitSetPresentationMode('picture-in-picture');
```
OrenMe: as this is sync assume success and set `_isInPictureInPicture` true, in catch set to false

OrenMe: in safari emit the player enter pip event ourselves, possibly make it async(if this is what chrome does) using setTimeout 0

src/engines/html5/html5.js
```diff
@@ -941,4 +947,57 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+      // Currently it's supported in chrome and in safari. So if we consider checking support before,
+      // we can use this flag to distinguish between two. In the future we might need a different method.
+      if (document.pictureInPictureEnabled) {
+        document
+          .exitPictureInPicture()
+          .catch(error =>
+            this.dispatchEvent(new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.CANNOT_EXIT_PICTURE_IN_PICTURE, error))
+          );
+      } else {
+        this._el.webkitSetPresentationMode('inline');
```
OrenMe: in safari emit the player leave pip event ourselves, possibly make it async(if this is what chrome does) using setTimeout 0

src/engines/html5/html5.js
```diff
@@ -941,4 +947,57 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+    }
+  }
+
+  exitPictureInPicture(): void {
+    try {
+      // Currently it's supported in chrome and in safari. So if we consider checking support before,
+      // we can use this flag to distinguish between two. In the future we might need a different method.
+      if (document.pictureInPictureEnabled) {
+        document
+          .exitPictureInPicture()
```
OrenMe: same as in enter pip, use the promise chain for updating the state

src/engines/html5/html5.js
```diff
@@ -941,4 +947,57 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+          .catch(error =>
+            this.dispatchEvent(new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.CANNOT_EXIT_PICTURE_IN_PICTURE, error))
+          );
+      } else {
+        this._el.webkitSetPresentationMode('inline');
+      }
+      this._isInPictureInPicture = false;
```
OrenMe: move this to the else for safari

src/engines/html5/html5.js
```diff
@@ -941,4 +941,89 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+      // Currently it's supported in chrome and in safari. So if we consider checking support before,
+      // we can use this flag to distinguish between the two. In the future we might need a different method.
+      // Second condition is because flow does not support this API yet
+      if (document.pictureInPictureEnabled && typeof this._el.requestPictureInPicture === 'function') {
+        this._el.requestPictureInPicture().catch(error => {
+          this.dispatchEvent(
+            new FakeEvent(
+              Html5EventType.ERROR,
+              new Error(Error.Severity.RECOVERABLE, Error.Category.PLAYER, Error.Code.PICTURE_IN_PICTURE_FAILURE, {
+                action: 'Enter picture in picture',
```
dan-ziv: why do we need the action property which is a free string?
why not to split error codes to `ENTER_PICTURE_IN_PICTURE_FAILED` & `EXIT_PICTURE_IN_PICTURE_FAILED`

src/event/event-type.js
```diff
@@ -90,7 +90,19 @@ const Html5EventType: PKEventTypes = {
...
+   */
+  ENTER_PICTURE_IN_PICTURE: 'enterpictureinpicture',
+  /**
+   * Fires when the engine exits picture in picture
+   */
+  LEAVE_PICTURE_IN_PICTURE: 'leavepictureinpicture',
+  /**
+   * Fires when the engine changes presentation mode on safari webkitpresentationmodechanged
+   */
+  WEBKIT_PRESENTATION_MODE_CHANGED: 'webkitpresentationmodechanged'
```
dan-ziv: remove the `WEBKIT` prefix since it is specific to chrome & safari. 
If FF will support this in the future it will be wrong to call it like that.

odedhutzler: changed

src/engines/html5/html5.js
odedhutzler: changed back to two errors.

PR #301, opened 2018-11-05
fix(FEC-8633): HLS fairplay fails with multiple players or consecutive playbacks

OrenMe:

The fairplay hander was managed as a static class and saved the media session on it.
When the player was reset the session wasn't closed and its reference remained on the static class.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/drm/fairplay.js
+++ b/src/drm/fairplay.js
@@ -1,6 +1,6 @@
 // @flow
 import BaseDrmProtocol from './base-drm-protocol';
-import type {fairplayDrmConfigType} from '../engines/html5/media-source/adapters/fairplay-drm-handler';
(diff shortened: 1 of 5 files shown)


src/drm/fairplay.js
```diff
@@ -1,10 +1,11 @@
 // @flow
 import BaseDrmProtocol from './base-drm-protocol';
-import Error from '../error/error';
+import type {fairplayDrmConfigType} from '../engines/html5/media-source/adapters/fairplay-drm-handler';
```
dan-ziv: Start type with upper case and move it to /flow-typed/types folder

OrenMe: Fixed

PR #311, opened 2018-11-21
fix(FEC-8712): replay button does not display when using a non liner ad

yairans:

wait to `ALL_ADS_COMPLETED` only when there is a post roll, else `ENDED` = `PLAYBACK_ENDED`

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/ads/ads-controller.js
+++ b/src/ads/ads-controller.js
@@ -2,15 +2,14 @@
 import Player from '../player';
 import {AdEventType} from './ad-event-type';
 import EventManager from '../event/event-manager';
-import FakeEventTarget from '../event/fake-event-target';
 import FakeEvent from '../event/fake-event';
 import {CustomEventType} from '../event/event-type';
 import Error from '../error/error';
 
-class AdsController extends FakeEventTarget implements IAdsController {
+class AdsController implements IAdsController {
   _player: Player;
   _adsPluginController: IAdsController;
-  _allAdsCompleted: boolean;
+  _hasPostRoll: boolean;
   _eventManager: EventManager;
 
   /**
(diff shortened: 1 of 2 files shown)


src/ads/ads-controller.js
```diff
@@ -19,19 +18,18 @@ class AdsController extends FakeEventTarget implements IAdsController {
...
 
   /**
-   * Getter to the no more ads to play indication.
-   * @returns {boolean} - Whether all ads completed.
+   * Getter whether there is a post roll
+   * @returns {boolean} - there is a post roll.
    */
-  get allAdsCompleted(): boolean {
-    return this._allAdsCompleted;
+  get hasPostRoll(): boolean {
```
dan-ziv: Dont forget this class is exposed via player.ads. thia API makes no sense to expose. App can get that information via the proper event payload.

dan-ziv: In addition, 
1. This is not a common ads API.
2. You broke ads API which is implemented also in the cast sender!

OrenMe: Ok, I tend to agree with @dan-ziv on this.
The api should remain allAdsCompletes and you can manage it internally by checking if there’s a postroll and if so wait for playback of post roll to update the status

yairans: @OrenMe @dan-ziv  I can't remain it as `allAdsCompletes` because now it returns `true` before `ALL_ADS_COMPLETED` event. consider case of preroll/midroll only.

yairans: @dan-ziv regarding the sender, i thought about this, but you anyway should your code, else you have same bug in case of non-linear ad.

OrenMe: @yairans still this API makes no real sense.
You can manage this internally.
allAdsCompletes can be set to true also in case of midroll cause after midroll it will be true.
The only thing this flag actually says is that there are no more ads.

yairans:
@dan-ziv this is was my initial solution but @OrenMe said that it's possible to be both linear and non-linear in same time

PR #316, opened 2018-11-25
fix(FEC-8722): more than one caption is checked at the caption menu

odedhutzler:

external text track will be active at the beginning in the following scenarios:
1. caption config has 'default: true' and playback.textLanguage is set to auto.
2. caption config has 'default: true' and playback.textLanguage is the same as the caption language.
otherwise, it will be set to false.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -2123,6 +2123,7 @@ export default class Player extends FakeEventTarget {
     const track: ?Track = this._getTracksByType(type).find(track => Track.langComparer(language, track.language));
     if (track) {
       this.selectTrack(track);
+      this._markActiveTrack(track);
     } else if (defaultTrack && !defaultTrack.active) {
       this.selectTrack(defaultTrack);
     }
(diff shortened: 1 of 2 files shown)


src/track/external-captions-handler.js
```diff
@@ -161,6 +161,9 @@ class ExternalCaptionsHandler extends FakeEventTarget {
    * @private
    */
   _createTextTrack(caption: PKExternalCaptionObject, index: number): TextTrack {
+    if (caption.default) {
+      this._isTextTrackActive = true;
```
OrenMe: should only be manipulated if set or hide are called

src/player.js
```diff
@@ -2123,6 +2123,7 @@ export default class Player extends FakeEventTarget {
     const track: ?Track = this._getTracksByType(type).find(track => Track.langComparer(language, track.language));
     if (track) {
       this.selectTrack(track);
+      this._markActiveTrack(track);
```
OrenMe: consider adding protection for only calling `this._markActiveTrack(track);` if `track.active`, because if it is not active we will call the engine and after it changes the track we will get a track change event and call  `_markActiveTrack` from there.

OrenMe: ok - so we now know we must call it as it registers some stuff in the engine - thanks @odedhutzler

PR #320, opened 2018-12-02
fix(FEC-8715): if label does not exist in manifest use the language property

odedhutzler:

if label does not exist - use the language property.

### CheckLists

- [x] Docs have been updated

### DIFF
--- a/src/track/text-track.js
+++ b/src/track/text-track.js
@@ -46,6 +46,8 @@ export default class TextTrack extends Track {
    */
   constructor(settings: Object = {}) {
     super(settings);
+    // some of the OTT dash manifests do not return a label, we will take the language instead.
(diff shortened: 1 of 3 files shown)


src/track/text-track.js
```diff
@@ -46,6 +46,8 @@ export default class TextTrack extends Track {
    */
   constructor(settings: Object = {}) {
     super(settings);
+    // some of the OTT dash manifests do not return a label, we will take the language instead.
```
itaykinnrot: remove the OTT - in general some dash will not have labels

OrenMe: don't think the comment should be here, not relevant in the OTT/OVP context.
You can just say: 

> use language tag if no display label is available

This is correct to any use case

PR #325, opened 2018-12-12
feat(FEC-8620): resize watcher

odedhutzler:

Added a resize watcher module to the player.
1. If the browser supports ResizeObserver (currently only chrome) than use it.
2. else - use an iframe of the size of the player.

### CheckLists

- [ ] Docs have been updated


src/utils/resize-watcher.js
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Start listening to a resize of the element.
+   * @param {string} playerId - the target to listen to.
+   * @returns {void}
+   */
+  init(playerId: string): void {
+    this._playerId = playerId;
+    window.ResizeObserver ? this._createNativeObserver() : this._createIframeObserver();
+    const el = document.getElementById(this._playerId);
```
OrenMe: Use utlis `getElementById` method
Maybe we need to pass the object already from player to library?

src/player.js
```diff
@@ -1581,9 +1586,8 @@ export default class Player extends FakeEventTarget {
       });
       this._eventManager.listen(this, CustomEventType.ENTER_FULLSCREEN, () => this._resetTextCuesAndReposition());
       this._eventManager.listen(this, CustomEventType.EXIT_FULLSCREEN, () => this._resetTextCuesAndReposition());
-      this._eventManager.listen(window, RESIZE, () => {
-        this._resetTextCuesAndReposition();
-      });
+      this._eventManager.listen(window, CustomEventType.RESIZE, () => this._resetTextCuesAndReposition());
```
OrenMe: do we still need this?

odedhutzler: nope

src/utils/resize-watcher.js
odedhutzler: yes i think thats better - it would be aligned with native / iframe observer api.

PR #326, opened 2018-12-12
fix(FEC-8764): after change media / refresh the video has no subtitles (hls)

odedhutzler:

since cues only exist in a 'hidden/showing' track and since all the other text tracks are disabled (by hls / or by the native adapter, [shaka has only 1 track]) then we can listen to all of them.

### CheckLists

- [ ] Docs have been updated


odedhutzler:
redundant.
Change the load promise to be returned only after "media attached", so the video track will already have mode='hidden', so it can be listened to.

PR #328, opened 2018-12-18
fix(FEC-8572): change the way the player compute text tracks font size

odedhutzler:

instead of using CSS class to override the vtt style, we will style each caption according to the settings.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -907,6 +907,19 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   }
 
   /**
+   * set hasBeenReset to true for all the cues. (use case: when cues should be recalculated for display)
+   * @returns {void}
+   */
+  resetAllCues(): void {
+    let activeTextTrack = Array.from(this._el.textTracks).find(track => track && track.mode !== 'disabled');
+    if (activeTextTrack) {
+      for (let i = 0; i < activeTextTrack.cues.length; i++) {
+        activeTextTrack.cues[i].hasBeenReset = true;
+      }
+    }
+  }
+
+  /**
    * Handles errors from the video element
    * @returns {void}
    * @private
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -326,6 +326,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     if (this._mediaSourceAdapter) {
       this._mediaSourceAdapter.selectTextTrack(textTrack);
     }
+    this.resetAllCues();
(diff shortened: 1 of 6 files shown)


src/player.js
```diff
@@ -1116,9 +1116,16 @@ export default class Player extends FakeEventTarget {
         sheet.insertRule(`video.${ENGINE_CLASS_NAME}::cue { ${style.toCSS()} }`, 0);
       }
       this._textStyle = style;
-      for (let i = 0; i < this._activeTextCues.length; i++) {
-        this._activeTextCues[i].hasBeenReset = true;
+      let videoEl = this.getVideoElement();
+      if (videoEl) {
+        let activeTextTrack = Array.from(videoEl.textTracks).find(track => track && track.mode !== 'disabled');
```
itaykinnrot: Need to move to engine level

odedhutzler: moved + added in playkit-flash

src/player.js
```diff
@@ -1114,11 +1114,14 @@ export default class Player extends FakeEventTarget {
 
     try {
       if (this._config.playback.useNativeTextTrack) {
-        sheet.insertRule(`#${this._playerId}  video.${ENGINE_CLASS_NAME}::cue { ${style.toCSS()} }`, 0);
-      } else {
-        sheet.insertRule(`#${this._playerId} .${SUBTITLES_CLASS_NAME} > div > div > div { ${style.toCSS()} }`, 0);
+        sheet.insertRule(`#${this._playerId}video.${ENGINE_CLASS_NAME}::cue { ${style.toCSS()} }`, 0);
       }
       this._textStyle = style;
+      if (this._engine) {
```
OrenMe: this should be in the else clause - we don't want to run it for self rendered text tracks

src/engines/html5/html5.js
```diff
@@ -907,6 +908,19 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     this.dispatchEvent(new FakeEvent(CustomEventType.TEXT_CUE_CHANGED, {cues: activeCues}));
   }
 
+  /**
+   * set hasBeenReset to true for all the cues. (use case: when cues should be recalculated for display)
+   * @returns {void}
+   */
+  resetAllCues(): void {
+    let activeTextTrack = Array.from(this._el.textTracks).find(track => track && track.mode !== 'disabled');
```
OrenMe: what about the tracks that already have cues but are now in disabled mode?

odedhutzler: they will be reset when the user will select the textTrack which they belong to

PR #334, opened 2019-01-09
feat(FEC-8749): expose IMA ad data

yairans:

expose `width`, `height` and `bitrate` on the `Ad` class

### CheckLists

- [ ] Docs have been updated


src/ads/ad.js
```diff
@@ -11,6 +11,9 @@ class Ad {
   _posterUrl: ?string;
   _skipOffset: ?number;
   _linear: ?boolean;
+  _width: number;
```
dan-ziv: Set default values since this is not supported on cast ads.

yairans: done

PR #337, opened 2019-01-13
fix(FEC-8763): add loadMedia life cycle hook for plugins

dan-ziv:

plugins life cycle should also have a life cycle method to identify that new media was loaded.
The new life cycle hook will enable returning event listeners or set up internal logic for a new media.  

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/plugin/plugin-manager.js
+++ b/src/plugin/plugin-manager.js
@@ -106,6 +106,15 @@ export default class PluginManager {
   }
 
   /**
+   * Iterates over all the plugins and calls loadMedia().
+   * @public
+   * @returns {void}
+   */
+  loadMedia(): void {
+    Object.keys(this._plugins).forEach(k => this._plugins[k].loadMedia());
+  }
+
+  /**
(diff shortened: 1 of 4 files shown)


src/plugin/plugin-manager.js
```diff
@@ -105,6 +105,15 @@ export default class PluginManager {
...
   }
 
+  /**
+   * Iterates over all the plugins and calls loadMedia().
+   * @public
+   * @returns {void}
+   */
+  loadMedia(): void {
+    Object.keys(this._plugins).forEach(k => this._plugins[k].loadMedia());
+  }
```
odedhutzler: why creating an array and then iterating and not just loop the object?

OrenMe: I think when we started doing object iterations the benchmark showed this is better, but I see now that `for..in` is now very fast: https://hackernoon.com/5-techniques-to-iterate-over-javascript-object-entries-and-their-performance-6602dcb708a8

PR #346, opened 2019-02-12
feat(FEC-8803): add 'playrequested' event

odedhutzler:
Adding `playrequested` event after a play was initiated (by the user or programmatically).

This will be used to calculate the `joinTime` (which was calculated from the play event sent from the video element until now)

### CheckLists

- [ ] Docs have been updated


OrenMe:
We will use `PLAYBACK_START`, so no need for this new event ATM

PR #357, opened 2019-04-15
FEC-9012 - We dont read mediaType VOD configuration

RoyBregman:
When we get in the configuration that mediatype is vod and the asset is live we return that the player is in live state - the actual result here should be that the player will load in VOD mode and not in Live mode (UI)

The change was to add a check that isLive in the player will return false if explicitly set as type VOD

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -948,7 +948,10 @@ export default class Player extends FakeEventTarget {
    * @public
    */
   isLive(): boolean {
-    return !!(this._config.sources.type === MediaType.LIVE || (this._engine && this._engine.isLive()));
+    return !!(
+      this._config.sources.type !== MediaType.VOD &&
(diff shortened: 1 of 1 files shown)


OrenMe:
@RoyBregman please pay attention to conventions. The title of the PR can’t be merged like this!
@yairans FYI

OrenMe:
Needs to be

> fix(FEC-9012): add mediaType VOD configuration check fonisLive verification

Reference for standard-version: https://github.com/conventional-changelog/standard-version#commit-message-convention-at-a-glance

RoyBregman:
fix(FEC-9012): add mediaType VOD configuration check for isLive verification

OrenMe:
@RoyBregman please delete the branch once merged(use the delete branch button in the PR)

PR #361, opened 2019-05-16
FEC-8975 QoS data enhancement

RoyBregman:

Added FRAG_LOAD and MANIFEST_LOAD events

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -302,6 +302,24 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   }
 
   /**
+   * @returns {number} - the number of video frames dropped since the creation of the associated HTMLVideoElement
+   */
+  getDroppedVideoFrames(): ?number {
+    if (this._droppedFramesWatcher) {
+      return this._droppedFramesWatcher._getDroppedAndDecodedFrames()[0];
+    }
+  }
+
+  /**
+   * @returns {number} - the number of video frames created and dropped since the creation of the associated HTMLVideoElement
+   */
+  getTotalVideoFrames(): ?number {
+    if (this._droppedFramesWatcher) {
+      return this._droppedFramesWatcher._getDroppedAndDecodedFrames()[1];
+    }
+  }
+
+  /**
    * Select a new video track.
    * @param {VideoTrack} videoTrack - The video track object to set.
    * @returns {void}
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -644,7 +644,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @public
    */
   get buffered(): TimeRanges {
-    return this._mediaSourceAdapter ? this._mediaSourceAdapter.buffered : 0;
+    return this._el.buffered;
   }
 
   /**
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -306,7 +306,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   getDroppedVideoFrames(): ?number {
     if (this._droppedFramesWatcher) {
-      return this._droppedFramesWatcher._getDroppedAndDecodedFrames()[0];
+      return this._droppedFramesWatcher.getDroppedAndDecodedFrames()[0];
     }
   }
 
@@ -315,7 +315,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   getTotalVideoFrames(): ?number {
     if (this._droppedFramesWatcher) {
-      return this._droppedFramesWatcher._getDroppedAndDecodedFrames()[1];
+      return this._droppedFramesWatcher.getDroppedAndDecodedFrames()[1];
     }
   }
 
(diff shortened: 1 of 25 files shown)


src/engines/html5/html5.js
```diff
@@ -299,6 +301,24 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     return this._el;
   }
 
+  /**
+   * @returns {number} - the number of video frames dropped since the creation of the associated HTMLVideoElement
+   */
+  getDroppedVideoFrames(): ?number {
+    if (this._droppedFramesWatcher) {
+      return this._droppedFramesWatcher._getDroppedAndDecodedFrames()[0];
```
yairans: `_getDroppedAndDecodedFrames` isa private method

src/engines/html5/html5.js
```diff
@@ -642,7 +644,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @public
    */
   get buffered(): TimeRanges {
-    return this._el.buffered;
+    return this._mediaSourceAdapter ? this._mediaSourceAdapter.buffered : 0;
```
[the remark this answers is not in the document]
RoyBregman: You are not looking at the latest code. This was removed later

src/engines/html5/media-source/base-media-source-adapter.js
```diff
@@ -264,4 +264,12 @@ export default class BaseMediaSourceAdapter extends FakeEventTarget implements I
   get capabilities(): PKMediaSourceCapabilities {
     return this._capabilities;
   }
+
+  get targetBuffer(): number {
+    return NaN;
+  }
+
+  get buffered(): TimeRanges {
```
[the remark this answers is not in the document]
RoyBregman: Same here

src/player.js
```diff
@@ -647,6 +647,26 @@ export default class Player extends FakeEventTarget {
...
+    const videoElement = this.getVideoElement();
+    let statsObject: PKStatsObject = {
+      targetBuffer: NaN,
+      availableBuffer: NaN
+    };
+    if (this._engine) {
+      statsObject.targetBuffer = this._engine.targetBuffer;
+      statsObject.availableBuffer = this._engine.availableBuffer;
+      // consideration of the end of the playback in the target buffer calc
+      if (!this.isLive()) {
```
OrenMe: I think you can simplify this to 
```javascript 
if (!this.isLive() && videoElement) {
  statsObject.targetBuffer = Math.min(statsObject.targetBuffer, videoElement.duration - videoElement.currentTime);
}
```
You can also move the `const videoElement = this.getVideoElement();`  inside the of as you only use it there 

Also, did you add the new engine members to flash and YouTube engines?

OrenMe: Also, shouldn’t this be `if ((!this.isLive() || this.isDvr()) && vidoeElement)`

src/player.js
```diff
@@ -647,6 +647,26 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  get stats(): PKStatsObject {
+    const videoElement = this.getVideoElement();
+    let statsObject: PKStatsObject = {
+      targetBuffer: NaN,
+      availableBuffer: NaN
+    };
+    if (this._engine) {
```
OrenMe: You can also substitute this with `if (videoElement)` and remove it from the bottom check as `this.getVideoElement()` validates for `_engine` as well so you can omit it here

src/player.js
RoyBregman: The dvr handling is being done in HLS and Dash as this depends on the player implementation

RoyBregman: > I think you can simplify this to
> 
> ```js
> if (!this.isLive() && videoElement) {
>   statsObject.targetBuffer = Math.min(statsObject.targetBuffer, videoElement.duration - videoElement.currentTime);
> }
> ```
> 
> You can also move the `const videoElement = this.getVideoElement();` inside the of as you only use it there
> 
> Also, did you add the new engine members to flash and YouTube engines?

Nope, needs to be done...

src/player.js
RoyBregman: I know it sounds weird, but can there be an engine but no video elements?

OrenMe: No, `this._el` in the engines, which is the response of `getVideoElement` is done in engine constructor.
Also, as mentioned, `this.getVideoElement()` is protected by itself with check fir `_engine`

src/player.js
```diff
@@ -657,10 +656,9 @@ export default class Player extends FakeEventTarget {
       statsObject.targetBuffer = this._engine.targetBuffer;
       statsObject.availableBuffer = this._engine.availableBuffer;
       // consideration of the end of the playback in the target buffer calc
-      if (!this.isLive()) {
-        statsObject.targetBuffer = videoElement
-          ? Math.min(statsObject.targetBuffer, videoElement.duration - videoElement.currentTime)
-          : statsObject.targetBuffer;
+      if (!this.isLive() && this.getVideoElement()) {
+        const videoElement = this.getVideoElement();
+        statsObject.targetBuffer = Math.min(statsObject.targetBuffer, videoElement.duration - videoElement.currentTime);
```
OrenMe: ok, I now understand this is not safe, as accessing the videoElement duration and currentTime is not guaranteed, as not all are HTML5 video element.
So I guess best place at the moment is in the base media source adapter in the `targetBuffer` getter

PR #364, opened 2019-05-27
feat(FEC-9055): Add custom Playback rate support

shalommeoded:

Exposed the Playback rate array, for user customization

### CheckLists

- [x] Docs have been updated

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -795,7 +795,7 @@ var config = {
 >
 > > ### config.playback.playbackRate
 > >
-> > ##### Type: `Array<Number>`
+> > ##### Type: `Array<number>`
 > >
 > > ##### Description: Sets the rate at which the media is being played back.
 > >
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -53,6 +53,7 @@ var config = {
     allowMutedAutoPlay: true,
     loop: false,
     muted: false,
+    playbackRate: [0.5, 1, 2, 4],
     pictureInPicture: true,
     options: {
       html5: {
@@ -790,6 +791,18 @@ var config = {
 > > ##### Description: Indicates whether the video should be muted or not.
 > >
 > > This is a Boolean attribute that indicates the default setting of the audio contained in the video. If set, the audio will be initially silenced. The attribute's default value is false, which means that the audio will be played automatically when the video is played.
+>
+> ##
+>
+> > ### config.playback.playbackRate
+> >
+> > ##### Type: `Array<Number>`
+> >
+> > ##### Default: `[0.5, 1, 2, 4]`
+> >
+> > ##### Description: Sets the rate at which the media is being played back.
+> >
+> > This is an Array attribute that is used to implement user controls for fast forward, slow motion, and so forth. The normal playback rate is multiplied by this value to obtain the current rate, so a value of 1.0 indicates normal speed.
 >
 > ##
 >
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -53,7 +53,6 @@ var config = {
     allowMutedAutoPlay: true,
     loop: false,
     muted: false,
-    playbackRate: [0.5, 1, 2, 4],
     pictureInPicture: true,
     options: {
       html5: {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -797,8 +797,6 @@ var config = {
 > >
 > > ##### Type: `Array<Number>`
 > >
-> > ##### Default: `[0.5, 1, 2, 4]`
-> >
 > > ##### Description: Sets the rate at which the media is being played back.
 > >
 > > This is an Array attribute that is used to implement user controls for fast forward, slow motion, and so forth. The normal playback rate is multiplied by this value to obtain the current rate, so a value of 1.0 indicates normal speed.
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -904,19 +904,15 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @return {number[]} - playback rates
    */
   get playbackRates(): Array<number> {
-    if (Array.isArray(this._config.playback.playbackRate)) {
-      const validPlaybackRateValue = this._config.playback.playbackRate.filter(number => {
-        if (typeof number === 'number' && number > 0) {
-          return number;
-        }
-      });
-      if (validPlaybackRateValue.length >= 1) {
-        return validPlaybackRateValue;
-      } else {
-        return Html5.PLAYBACK_RATES;
-      }
-    }
-    return Html5.PLAYBACK_RATES;
+    return this._config.playbackRates;
+  }
+
+  /**
+   * set the playback rates
+   * @param {number[]} playbackRates - playback rates
+   */
+  set playbackRates(playbackRates: Array<number>): void {
+    this._config.playbackRates = playbackRates;
   }
 
   /**
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -904,6 +904,18 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @return {number[]} - playback rates
    */
   get playbackRates(): Array<number> {
+    if (Array.isArray(this._config.playback.playbackRate)) {
+      const validPlaybackRateValue = this._config.playback.playbackRate.filter(number => {
+        if (typeof number === 'number' && number > 0) {
+          return number;
+        }
+      });
+      if (validPlaybackRateValue.length >= 1) {
+        return validPlaybackRateValue;
+      } else {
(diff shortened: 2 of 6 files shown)


src/engines/html5/html5.js
```diff
@@ -904,6 +904,18 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @return {number[]} - playback rates
    */
   get playbackRates(): Array<number> {
+    if (Array.isArray(this._config.playback.playbackRate)) {
```
OrenMe: this means every time the API is called we will have to do this check. Sounds a bit redundant.
Maybe add a setter and do the validation there? @yairans WDYT?

src/player-config.js
```diff
@@ -24,6 +24,7 @@ const DefaultConfig = {
     loop: false,
     allowMutedAutoPlay: true,
     muted: false,
+    playbackRate: [0.5, 1, 2, 4],
```
OrenMe: This is not correct, the config needs to be taken from the engine as only it knows what he can support.
Adding a default will cause the player to return this instead of engine supported playback rates

src/engines/html5/html5.js
```diff
@@ -904,6 +904,18 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+      const validPlaybackRateValue = this._config.playback.playbackRate.filter(number => {
+        if (typeof number === 'number' && number > 0) {
```
yairans:
```suggestion
        if (number > 0) {
```

src/engines/html5/html5.js
```diff
@@ -904,6 +904,18 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @return {number[]} - playback rates
    */
   get playbackRates(): Array<number> {
+    if (Array.isArray(this._config.playback.playbackRate)) {
+      const validPlaybackRateValue = this._config.playback.playbackRate.filter(number => {
```
yairans:
```suggestion
      const validPlaybackRateValue = this._config.playback.playbackRate.filter(number => (number > 0));
```

src/engines/html5/html5.js
yairans: agree

docs/configuration.md
```diff
@@ -53,6 +53,7 @@ var config = {
     allowMutedAutoPlay: true,
     loop: false,
     muted: false,
+    playbackRate: [0.5, 1, 2, 4],
```
yairans: remove. this depends on the engine 
also below in the description

src/player.js
```diff
@@ -1731,6 +1741,12 @@ export default class Player extends FakeEventTarget {
     if (typeof this._config.playback.crossOrigin === 'string') {
       this.crossOrigin = this._config.playback.crossOrigin;
     }
+    if (Array.isArray(this._config.playback.playbackRate)) {
+      const validPlaybackRateValue = this._config.playback.playbackRate.filter(number => number > 0 && number <= 16);
```
yairans: name should be `validPlaybackRates`. `filter` returns an array.

src/engines/html5/html5.js
```diff
@@ -904,7 +904,15 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
-    return Html5.PLAYBACK_RATES;
+    return this._config.playbackRates;
+  }
+
+  /**
+   * set the playback rates
+   * @param {number[]} playbackRates - playback rates
+   */
+  set playbackRates(playbackRates: Array<number>): void {
+    this._config.playbackRates = playbackRates;
```
yairans: html should return always `Html5.PLAYBACK_RATES` the overriding should be handled in the player playbackRates getter:
```
if (this._playbackRates) {
  return this._playbackRates;
} else if (this._engine) {
  return this._engine.playbackRates;
}
return [];
```
`this._playbackRates` should be set in `_handlePlaybackOptions`

src/player.js
```diff
@@ -1731,6 +1743,13 @@ export default class Player extends FakeEventTarget {
     if (typeof this._config.playback.crossOrigin === 'string') {
       this.crossOrigin = this._config.playback.crossOrigin;
     }
+    if (Array.isArray(this._config.playback.playbackRate)) {
+      const validPlaybackRates = this._config.playback.playbackRate.filter(number => number > 0 && number <= 16);
+      if (validPlaybackRates) {
+        this._playbackRates = validPlaybackRates;
+      }
+      this._playbackRates = this._engine.PlaybackRates;
```
yairans: remove. `this._engine` could be null here

src/player.js
```diff
@@ -844,12 +844,24 @@ export default class Player extends FakeEventTarget {
...
       return this._engine.playbackRates;
     }
     return [];
   }
 
+  /**
+   * Sets the playbackRates property.
+   * @param {number} playbackRates - Set the possible playback speeds speed of the video.
+   */
+  set playbackRates(playbackRates: number): void {
```
yairans: where's this using?

src/player.js
```diff
@@ -820,22 +820,22 @@ export default class Player extends FakeEventTarget {
   }
 
   /**
-   * Sets the playbackRate property.
-   * @param {number} rate - The playback speed of the video.
+   * Gets the current playback speed of the video.
```
yairans: did u change the set/get order? revert

docs/configuration.md
```diff
@@ -793,6 +793,18 @@ var config = {
 >
 > ##
 >
+> > ### config.playback.playbackRate
+> >
+> > ##### Type: `Array<Number>`
+> >
+> > ##### Default: `[0.5, 1, 2, 4]`
```
yairans: depends on the engine

src/player.js
shalommeoded: Forgot that `_handlePlaybackOptions` removed.

src/player.js
shalommeoded: Removed

docs/configuration.md
shalommeoded: Removed

src/player-config.js
yairans: what's about this?

shalommeoded:
@yairans 

Added a test, also added an additional use case if we have a duplicate number.

test/src/player.spec.js
```diff
@@ -2105,6 +2105,13 @@ describe('Player', function() {
       player.config.playback.streamPriority = {};
       player.config.playback.streamPriority.should.deep.equal(getConfigStructure().playback.streamPriority);
     });
+
+    it('should set playbackRate', function() {
+      config.playback.playbackRate.concat([-1, 2, 4, 3, 20, 3, 2]);
+      player = new Player(config);
+      playerContainer.appendChild(player.getView());
+      player.config.playback.playbackRate.should.deep.equal(getConfigStructure().playback.playbackRate);
```
OrenMe: if I'm not missing anything, a proper test will be to actually compare player.playbackRates to the config passed to it

src/player.js
```diff
@@ -1731,6 +1733,14 @@ export default class Player extends FakeEventTarget {
     if (typeof this._config.playback.crossOrigin === 'string') {
       this.crossOrigin = this._config.playback.crossOrigin;
     }
+    if (Array.isArray(this._config.playback.playbackRate)) {
+      const validPlaybackRates = this._config.playback.playbackRate.filter(
+        (number, index, self) => number > 0 && number <= 16 && self.indexOf(number) === index
+      );
```
OrenMe: add sorting `.sort((a, b) => a-b)`

shalommeoded: @OrenMe 
Added

test/src/player.spec.js
OrenMe: @shalommeoded ping about the test

shalommeoded: @OrenMe 
done

src/player.js
```diff
@@ -844,7 +844,9 @@ export default class Player extends FakeEventTarget {
    * @returns {Array<number>} - The possible playback speeds speed of the video.
    */
   get playbackRates(): Array<number> {
-    if (this._engine) {
+    if (this._playbackRates) {
```
OrenMe: need to define `_playbackRates: Array<numbers>`

shalommeoded: done

src/player.js
```diff
@@ -245,6 +245,12 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _playbackStart: boolean;
+  /**
+   * The available playback rates for the player.
+   * @type {Array<numbers>}
+   * @private
+   */
+  _playbackRates: Array<numbers>;
```
OrenMe: no such thing `numbers`, `Array<number>`, also in JSDoc
Also in the `.md` file

shalommeoded: @OrenMe 
Fixed and checked with yarn flow as well.

PR #372, opened 2019-07-03
feat(FEC-9227): support reInit of MSE

Yuvalke:

Destroy and initilize hls.js before and after ads for now(should be resolved by hls.js later and it'll be replace by attach detach media)

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -444,6 +444,7 @@ export default class Player extends FakeEventTarget {
       this._configureOrLoadPlugins(config.plugins);
       this._maybeCreateAdsController();
       this.reset();
+      this.handleAdsWithMSE();
       this._resizeWatcher.init(Utils.Dom.getElementById(this._playerId));
       Player._logger.debug('Change source started');
       this.dispatchEvent(new FakeEvent(CustomEventType.CHANGE_SOURCE_STARTED));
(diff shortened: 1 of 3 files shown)


src/player.js
```diff
@@ -635,7 +636,36 @@ export default class Player extends FakeEventTarget {
...
+      this._engine.detachMediaSource();
+    }
+  }
+  handleAdsWithMSE(): void {
+    if (this.config.playback.playAdsWithMSE) {
+      this._eventManager.listen(this, AdEventType.AD_BREAK_START, this.detachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_BREAK_END, this.attachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_ERROR, this.attachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_SKIPPED, this.attachMediaSource);
+      this._eventManager.listen(this, AdEventType.USER_CLOSED_AD, this.attachMediaSource);
```
OrenMe: this is not relevant - it is for non linear ads that don't use the video

src/player.js
```diff
@@ -635,7 +636,36 @@ export default class Player extends FakeEventTarget {
...
+    if (this._engine) {
+      this._engine.detachMediaSource();
+    }
+  }
+  handleAdsWithMSE(): void {
+    if (this.config.playback.playAdsWithMSE) {
+      this._eventManager.listen(this, AdEventType.AD_BREAK_START, this.detachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_BREAK_END, this.attachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_ERROR, this.attachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_SKIPPED, this.attachMediaSource);
```
OrenMe: ad skip should result in `AD_BREAK_END` so not necessary

src/player.js
```diff
@@ -635,7 +636,36 @@ export default class Player extends FakeEventTarget {
...
+  detachMediaSource(): void {
+    if (this._engine) {
+      this._engine.detachMediaSource();
+    }
+  }
+  handleAdsWithMSE(): void {
+    if (this.config.playback.playAdsWithMSE) {
+      this._eventManager.listen(this, AdEventType.AD_BREAK_START, this.detachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_BREAK_END, this.attachMediaSource);
+      this._eventManager.listen(this, AdEventType.AD_ERROR, this.attachMediaSource);
```
OrenMe: need to better understand this, in general `AD_BREAK_END` should be fired if ad was loaded and then failed.
Maybe you are referring to a case where the initial ad request failed and video ad never loaded(i.e. also `AD_BREAK_START` never happened) and in this case is this tested and you know this is needed?

src/player.js
```diff
@@ -635,7 +636,36 @@ export default class Player extends FakeEventTarget {
...
+   * Listen to the video element events and triggers them from the engine.
+   * @public
+   * @returns {void}
+   */
+  detachMediaSource(): void {
+    if (this._engine) {
+      this._engine.detachMediaSource();
+    }
+  }
+  handleAdsWithMSE(): void {
```
OrenMe: is this a public API?

src/player.js
```diff
@@ -444,6 +444,7 @@ export default class Player extends FakeEventTarget {
       this._configureOrLoadPlugins(config.plugins);
       this._maybeCreateAdsController();
       this.reset();
+      this.handleAdsWithMSE();
```
OrenMe: this seems like a player instance level config, and not source level config, so this is not a proper location

PR #374, opened 2019-07-08
Fec 8696 contrib presets

esakal:


src/player.js
```diff
@@ -390,6 +390,12 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _resizeWatcher: ResizeWatcher;
+  /**
+   * Holds preset component factories
+   * @type {?PresetComponent}
+   * @private
+   */
+  _presetComponents: PresetComponent[];
```
yairans:
```suggestion
  _presetComponents: Array<PresetComponent>;
```
fix the the jsdoc accordingally

src/player.js
```diff
@@ -916,6 +923,10 @@ export default class Player extends FakeEventTarget {
     return Utils.Object.mergeDeep({}, this._config);
   }
 
+  get presetComponents(): PresetComponent[] {
```
yairans:
```suggestion
  get presetComponents(): Array<PresetComponent> {
```

src/player.js
```diff
@@ -916,6 +923,10 @@ export default class Player extends FakeEventTarget {
...
+    return [...this._presetComponents];
```
yairans:
```suggestion
    return this._presetComponents;
```

src/player.js
```diff
@@ -1340,6 +1351,15 @@ export default class Player extends FakeEventTarget {
     setLogLevel(level, name);
   }
 
+  /**
+   * get new logger
+   * @param {*} name name
+   * @returns {void}
+   */
+  getLogger(name: string): void {
```
yairans: why this new api?

src/player.js
```diff
@@ -1488,12 +1509,17 @@ export default class Player extends FakeEventTarget {
               if (typeof plugin.getMiddlewareImpl === 'function') {
                 plugin.name === 'bumper' ? middlewares.push(plugin.getMiddlewareImpl()) : middlewares.unshift(plugin.getMiddlewareImpl());
               }
+
+              if (typeof plugin.getPresetComponents === 'function') {
+                presetComponents.push(...(plugin.getPresetComponents() || []));
```
yairans:
```suggestion
                presetComponents.concat(plugin.getPresetComponents() || []);
```

flow-typed/types/preset.types.js
```diff
@@ -0,0 +1,11 @@
+
+type PresetComponent = {
```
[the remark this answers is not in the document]
yairans: I'm for Option 1. The hierarchy structure is easier to understand.

kaltura-hooks:
Hi @esakal,

flow-typed/types/preset.types.js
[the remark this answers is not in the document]
OrenMe: what are the pros/cons of each, can you please explain?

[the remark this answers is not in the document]
esakal: @yairnas can you please reply to Oren?

esakal:
@OrenMe @yairans updated pr per your request

src/player.js
esakal: removed

src/utils/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './util';
+export {ResizeWatcher} from './resize-watcher';
```
OrenMe: ResizeWatcher is an internal utility, and is exposed via event to app, we don't plan to expose it for actual usage a utility.

src/utils/resize-watcher.js
```diff
@@ -19,6 +19,7 @@ class ResizeWatcher extends FakeEventTarget {
    * @returns {void}
    */
   destroy(): void {
+    // todo sakal Oren I think it should also remove all listeners, you might have memory leaks
```
OrenMe: disconnect does unobserv and clears watcher: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/disconnect
For iframe api we clear the `onresize` ourselves.
Which event listener is left?

src/assets/style.css
```diff
@@ -61,7 +61,6 @@
   width: 100%;
   height: 100%;
   position: absolute;
-  visibility: hidden;
   z-index: -1;
```
OrenMe: let's remove the border as well, and set bigger negative z-index for safety.
Good catch!

src/utils/index.js
esakal: I need this one to be used in playkit-js-ui to support side panels correctly, I will explain in our meeting.

src/utils/resize-watcher.js
esakal: I might be wrong here, this class inherit from `FakeEventTarget` which handles its own queue of listeners. So while you unobserve the window you still hold reference to handler functions of components.

src/utils/index.js
OrenMe: OK, but I would advise you have a contingency plan as I don't see a reason to use more than one in the player core, and I see you already built upon it in UI manager.

src/utils/resize-watcher.js
OrenMe: `FakeEventTarget` is the base object, it mimics the native DOM event target.
It is used by an event manager to register and unregister from listeners.
It is the responsibility of the listener invocation to unregister, which is done in player.js.

OrenMe:
done in #387

PR #380, opened 2019-07-23
fix(FEC-9261): playback doesn't play with non-linear ads

Yuvalke:

add check for non linear ads for attachMediaSource/detachMediaSource, shouldn't use attachMediaSource/detachMediaSource with non-linear ads

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -646,10 +646,11 @@ export default class Player extends FakeEventTarget {
   /**
    * detach the engine's media source
    * @private
+   * @param {FakeEvent} event - loaded event
    * @returns {void}
    */
-  _detachMediaSource(): void {
-    if (this._engine) {
+  _detachMediaSource(event: FakeEvent): void {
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -646,10 +646,11 @@ export default class Player extends FakeEventTarget {
   /**
    * detach the engine's media source
    * @private
+   * @param {FakeEvent} event - loaded event
    * @returns {void}
    */
-  _detachMediaSource(): void {
-    if (this._engine) {
+  _detachMediaSource(event: FakeEvent): void {
```
OrenMe: this method shouldn't care about the reason why to do the detach and we should keep it clear of this logic.
Move this to where the context of decision is the ad, the `AD_LOADED` event listener 
https://github.com/kaltura/playkit-js/blob/b50bf983fb2f32b3d5f39ea309d0d3eddf3afb95/src/player.js#L1681-L1685

PR #381, opened 2019-07-24
fix(FEC-9262): playback doesn't return to start after playback with ads in LG TV

Yuvalke:

Add support for replay after ads finish to return to beginning. 
fix bug engine.ended on `attachMediaSource` will not give always true on the end of the media as expected, instead on `detachMediaSource` which will return true when ad playing and decorator is active or when decorator is not active engine will return true from video element.
keeps the value form `detachMediaSource` to know in `attachMediaSource` if media ended.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -46,12 +46,6 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   _config: Object;
   /**
-   * The player ended
-   * @type {boolean}
-   * @private
-   */
-  _playbackEndedOnDetach: boolean = false;
-  /**
    * Promise to indicate when a media source adapter can be loaded.
    * @type {Promise<*>}
    * @private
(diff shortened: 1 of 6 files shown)


src/engines/html5/html5.js
```diff
@@ -253,17 +253,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   attachMediaSource(): void {
     if (this._mediaSourceAdapter) {
-      //added to mask problematic behavior in shaka - init fire loadstart event, only for last init to avoid spinner
       this._mediaSourceAdapter.attachMediaSource(this._playbackEndedOnDetach);
-      if (this._playbackEndedOnDetach) {
```
OrenMe: @Yuvalke so what is the solution here if you remove this?

Yuvalke: Shaka init on load, so i didn't init on attaching cause when load call it'll be init again.
On the end load doesn't happen

OrenMe: Good news, so we can wrap this with the other PRs or anything else is left?

Yuvalke: I have a small fix which will remove the parameters, there are not needed anymore.
Every adapter can handle every by himeself

Yuvalke: I have a small fix which will remove the parameters, there are not needed anymore.
Every adapter can handle every by himeself

PR #382, opened 2019-07-25
fix(FEC-8873): iPhone raises error when coming back from idle

RoyBregman:
reloading the player and bring the video to the same state so user will not notice the error


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,10 +281,50 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+        }
+      });
+
+      this._videoElement.play();
+    });
+  }
+
   handleMediaError(error: ?MediaError): boolean {
     if (this._loadPromiseReject) {
+      NativeAdapter._logger.debug('handleMediaError _loadPromiseReject');
```
OrenMe: please remove, we have printouts for error event in console

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,10 +281,50 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
       this._loadPromiseReject(new Error(Error.Severity.CRITICAL, Error.Category.MEDIA, Error.Code.NATIVE_ADAPTER_LOAD_FAILED, error));
       return true;
+      // It seems that the only known scenario to reach this code with a decode error (3) is as described in FEC-8873
+    } else if (error && error.code === 3) {
```
OrenMe: need to add retry params so we don't try to recover and fail forever.
Use platform ENUM instead of 3, i.e. `MediaError.MEDIA_ERR_DECODE`, see https://developer.mozilla.org/en-US/docs/Web/API/MediaError

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -669,9 +711,11 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     };
     NativeAdapter._logger.debug('Video element audio track change');
     const vidIndex = getActiveVidAudioTrackIndex();
-    const pkIndex = getActivePKAudioTrackIndex();
+    const activeAudioTrack = this._getActivePKAudioTrack();
```
OrenMe: if this can return null/undefined then you have an issue with the types in `_getActivePKAudioTrack`

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -725,18 +769,24 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     }
   }
 
+  _getPKTextTracks(): AudioTrack[] {
```
OrenMe: wrong type

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -725,18 +769,24 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+    return this._playerTracks.filter(track => track instanceof PKTextTrack);
+  }
+
+  _getActivePKTextTrack(): AudioTrack {
```
OrenMe: wrong type

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -748,7 +798,8 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     };
     NativeAdapter._logger.debug('Video element text track change');
     const vidIndex = getActiveVidTextTrackIndex();
-    const pkIndex = getActivePKTextTrackIndex();
+    const activePKtextTrack = this._getActivePKTextTrack();
```
OrenMe: same as for audio, this may return null/undefined

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -855,6 +906,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
    * @returns {void}
    */
   _disableAudioTracks(): void {
+    NativeAdapter._logger.debug('_disableAudioTracks');
```
OrenMe: please remove

src/engines/html5/html5.js
```diff
@@ -1055,6 +1055,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @private
    */
   _handleVideoError(): void {
+    Html5._logger.debug('_handleVideoError');
```
OrenMe: please remove

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,10 +281,50 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+    NativeAdapter._logger.debug(`handleDecodeError currentTime <${this._videoElement.currentTime}>`);
+    this._playerTracks = this._getParsedTracks();
+    const prevCurrTime = this._videoElement.currentTime;
+    const prevActiveAudioTrack = this._getActivePKAudioTrack();
+    const prevActiveTextTrack = this._getActivePKTextTrack();
+
+    this._eventManager.removeAll();
+    this._loadPromise = null;
+
+    if (this._videoElement && this._videoElement.src) {
```
OrenMe: is this required? why?

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,10 +281,50 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     return this._loadPromise;
   }
 
+  /**
+   * handle decode error - reload the video and seek to last currentTime
+   * @returns {void}
+   */
+  handleDecodeError(): void {
+    NativeAdapter._logger.debug(`handleDecodeError currentTime <${this._videoElement.currentTime}>`);
+    this._playerTracks = this._getParsedTracks();
```
OrenMe: why do you need to get `this._getParsedTracks()` again?

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,10 +281,50 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+    const prevCurrTime = this._videoElement.currentTime;
+    const prevActiveAudioTrack = this._getActivePKAudioTrack();
+    const prevActiveTextTrack = this._getActivePKTextTrack();
+
+    this._eventManager.removeAll();
+    this._loadPromise = null;
```
OrenMe: don't you need to reset more members?
```javascript
this._waitingEventTriggered = false;
this._liveEdge = 0;
this._clearHeartbeatTimeout();
if (this._liveDurationChangeInterval) {
  clearInterval(this._liveDurationChangeInterval);
  this._liveDurationChangeInterval = null;
}
```

What about the DRM handler? if you reset a DRM playback does it recover correctly? need to check this

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,10 +281,50 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+
+    if (this._videoElement && this._videoElement.src) {
+      Utils.Dom.setAttribute(this._videoElement, 'src', '');
+      Utils.Dom.removeAttribute(this._videoElement, 'src');
+    }
+
+    this.load(prevCurrTime).then(() => {
+      this._eventManager.listenOnce(this._videoElement, Html5EventType.PLAYING, () => {
```
OrenMe: `load` is resolved when there's already track data and after `dataloaded` event so I'm not sure we must wait for playing.

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,10 +281,50 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+        if (prevActiveAudioTrack) {
```
OrenMe: is there a chance this is not always available? there is always an active audio(not like text track for example, where it could be actually off)

OrenMe:
fix in #390

PR #389, opened 2019-08-07
fix(FEC-8873): fatal error occurs when coming back from idle in Native HLS IOS

RoyBregman:
created a workaround that restores the player to previous state so the user will not "feel" the error


OrenMe:
solved in #390

PR #390, opened 2019-08-15
fix(FEC-8873): fatal error occurs when coming back from idle in Native HLS IOS

RoyBregman:
added recovery to continue playback from the same point (code effects only native-adapter)

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -15,6 +15,7 @@ import {FairplayDrmHandler} from './fairplay-drm-handler';
 import type {FairplayDrmConfigType} from './fairplay-drm-handler';
 
 const BACK_TO_FOCUS_TIMEOUT: number = 1000;
+const MAX_MEDIA_RECOVERY_ATTEMPTS: number = 3;
 
 /**
  * An illustration of media source extension for progressive download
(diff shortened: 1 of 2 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,13 +289,47 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
   handleMediaError(error: ?MediaError): boolean {
     if (this._loadPromiseReject) {
       this._loadPromiseReject(new Error(Error.Severity.CRITICAL, Error.Category.MEDIA, Error.Code.NATIVE_ADAPTER_LOAD_FAILED, error));
       return true;
-    } else {
-      return false;
+    } else if (error && error.code === window.MediaError.MEDIA_ERR_DECODE) {
+      this._mediaErrorRecoveryAttempts++;
+      if (this._mediaErrorRecoveryAttempts <= MAX_MEDIA_RECOVERY_ATTEMPTS) {
+        this.handleDecodeError();
```
OrenMe: pass the error object and print it in the `handleDecodeError`

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -281,13 +289,47 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     return this._loadPromise;
   }
 
+  /**
+   * handle decode error - reload the video and seek to last currentTime
+   * @returns {void}
+   */
+  handleDecodeError(): void {
```
OrenMe: make this private
```suggestion
  _handleDecodeError(): void {
```

Also change the jsdoc to mention private and the call from `handleMediaError`

PR #400, opened 2019-09-19
fix(FEC-8821): css fullscreen doesn't work with youtube iframe

Yuvalke:

object-fit: contain doesn't work as excepted on IOS iframe

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -21,6 +21,10 @@
   object-fit: contain;
 }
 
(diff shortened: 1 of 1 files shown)


src/assets/style.css
```diff
@@ -21,6 +21,10 @@
   object-fit: contain;
 }
 
+.playkit-in-browser-fullscreen-mode .playkit-engine-youtube {
```
OrenMe: if this is engine specific then add it in youtube engine

PR #409, opened 2019-10-22
fix(FEC-9379): add detection for chromium Edge

Yuvalke:

add user detection for dev or canary Chromium Edge

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/utils/env.js
+++ b/src/utils/env.js
@@ -20,11 +20,11 @@ const DeviceParser = [
   [[UAParser.DEVICE.TYPE, UAParser.DEVICE.SMARTTV]]
 ];
 
-const EdgeChromium = [[/(edg)\/((\d+)?[\w.]+)/i], [[UAParser.BROWSER.NAME, 'Edge'], UAParser.BROWSER.VERSION, UAParser.BROWSER.MAJOR]];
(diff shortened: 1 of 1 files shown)


src/utils/env.js
```diff
@@ -20,7 +20,11 @@ const DeviceParser = [
   [[UAParser.DEVICE.TYPE, UAParser.DEVICE.SMARTTV]]
 ];
 
-let Env = new UAParser(undefined, {browser: SAMSUNGBrowserParser, device: DeviceParser, os: LGOSParser}).getResult();
+const EdgeChromium = [[/(edg)\/((\d+)?[\w.]+)/i], [[UAParser.BROWSER.NAME, 'Edge'], UAParser.BROWSER.VERSION, UAParser.BROWSER.MAJOR]];
+
+const browserDetect = EdgeChromium.concat(SAMSUNGBrowserParser);
```
RoyBregman: is browserDetect the best naming for this variable? maybe something like samsungOrChromiumDetect?

Yuvalke: I don't think so, according to your suggestion next time it'll be samsungOrChromiumSomethingElseDetect and after that.. bottom line you don't want to create infinite variable names which will be meaningless,  I used the the name of the attribute that I want to parse.

RoyBregman: ok, so maybe just uppercase cause its a constant

Yuvalke: About the constant, it's `Edge` cause in chromium it wrote `Edg` in user agent and in other cases is `Edge`.

RoyBregman: my bad, I am used that constants are uppercased. But not in Kaltura :)

PR #410, opened 2019-10-23
fix(FEC-9427): use response type in xhr arraybuffer for Fairplay request

Yuvalke:

To be compatible with different DRM server
Solves FEC-9427

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/fairplay-drm-handler.js
+++ b/src/engines/html5/media-source/adapters/fairplay-drm-handler.js
@@ -34,7 +34,9 @@ class FairplayDrmHandler {
       responseFilter: (type, response) => {
         let responseObj = {};
         try {
-          let keyText = response.data.trim();
+          const dataView = new DataView(response.data);
+          const decoder = new TextDecoder();
+          const keyText = decoder.decode(dataView);
           responseObj = JSON.parse(keyText);
         } catch (error) {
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/fairplay-drm-handler.js
```diff
@@ -34,7 +34,9 @@ class FairplayDrmHandler {
       responseFilter: (type, response) => {
         let responseObj = {};
         try {
-          let keyText = response.data.trim();
```
OrenMe: should keep the trim in case server returns trailing spaces - or check with uDRM team if this is not necessary anymore

Yuvalke: Right, better to keep it anyway it'll safe

src/engines/html5/media-source/adapters/fairplay-drm-handler.js
```diff
@@ -34,7 +34,7 @@ class FairplayDrmHandler {
       responseFilter: (type, response) => {
         let responseObj = {};
         try {
-          const dataView = new DataView(response.data);
+          const dataView = new DataView(response.data.trim());
```
OrenMe: trim should be on the text output not the binary.

Yuvalke: trim wouldn't delete empty chars from binary?

PR #415, opened 2019-11-10
feat: new USER_GESTURE event and api

yairans:

* `USER_GESTURE` event
* `get hasUserInteracted` api

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -399,6 +399,12 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _uiComponents: Array<PKUIComponent>;
+  /**
+   * Whether the user interacted with the player
+   * @type {boolean}
+   * @private
+   */
+  _hasUserInteraction: boolean = false;
 
   /**
    * @param {Object} config - The configuration for the player instance.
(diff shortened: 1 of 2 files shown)


src/player.js
```diff
@@ -950,6 +956,15 @@ export default class Player extends FakeEventTarget {
     return [...this._uiComponents];
   }
 
+  /**
+   * Get whether the user already interacted with the player
+   * @returns {boolean} - Whether the user interacted with the player
+   * @public
+   */
+  get hasUserInteraction(): boolean {
```
RoyBregman: consider renaming to hasUserInteracted instead

src/player.js
```diff
@@ -1706,6 +1721,18 @@ export default class Player extends FakeEventTarget {
         this._eventManager.listen(this, AdEventType.AD_BREAK_END, this._attachMediaSource);
         this._eventManager.listen(this, AdEventType.AD_ERROR, this._attachMediaSource);
       }
+      const rootElement = Utils.Dom.getElementBySelector(`#${this.config.targetId}`);
```
Yuvalke: Could be an issue when player is smaller than root element.
Maybe playkit-video-player could do the trick

yairans: I tested it with a player of 50% - works fine

PR #422, opened 2019-11-28
fix(FEC-9471): slider progress bar exceeds 100%

RoyBregman:

Added boolean that holds the state of isOnLive edge
set to true when live video starts or when seeked to live or when currenttime is equal or greater of duration.
set to false when video is paused or when seeked to position smaller then duration

solves FEC-9471

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -719,9 +719,6 @@ export default class Player extends FakeEventTarget {
         }
         this._engine.currentTime = boundedTo;
       }
-      if (this.isLive()) {
-        this._isOnLiveEdge = this.duration ? to >= this.duration && !this.paused : false;
-      }
     }
   }
 
@@ -1672,6 +1669,11 @@ export default class Player extends FakeEventTarget {
           return this.dispatchEvent(event);
         });
       });
+      this._eventManager.listen(this._engine, Html5EventType.SEEKING, () => {
+        if (this.isLive()) {
+          this._isOnLiveEdge = this.duration && this.currentTime ? this.currentTime >= this.duration - 1 && !this.paused : false;
+        }
+      });
       this._eventManager.listen(this._engine, Html5EventType.SEEKED, () => {
         const browser = this._env.browser.name;
         if (browser === 'Edge' || browser === 'IE') {
@@ -1969,7 +1971,7 @@ export default class Player extends FakeEventTarget {
       this._engine
         .load(startTime)
         .then(data => {
-          if (this.isLive()) {
+          if (this.isLive() && startTime === -1) {
             this._isOnLiveEdge = true;
           }
           this._updateTracks(data.tracks);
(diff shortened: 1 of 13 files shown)


src/player.js
```diff
@@ -1949,6 +1969,9 @@ export default class Player extends FakeEventTarget {
       this._engine
         .load(startTime)
         .then(data => {
+          if (this.isLive()) {
```
Yuvalke: After load with explicit startTime you'll not on live edge

OrenMe: It is "safe" to assume that if someone set start time for live he will want to be in DVR and not live.
The default behaviour is to be on lice edge when loading a live stream without a start time.

Yuvalke: Live doesn't include DVR? I mean isLive will return true for DVR , right?

OrenMe: Live will return true for live with and without DVR.

Yuvalke: OK, the whole fix is about DVR cause seekbar doesn't exist for live, so if you start with start time you're not in the live edge but set liveEdge to true

OrenMe: was there a question there? :-)

Yuvalke: Nop, just explanation for the problematic scenario :)

src/player.js
```diff
@@ -1653,6 +1669,11 @@ export default class Player extends FakeEventTarget {
           return this.dispatchEvent(event);
         });
       });
+      this._eventManager.listen(this._engine, Html5EventType.SEEKING, () => {
+        if (this.isLive()) {
+          this._isOnLiveEdge = this.duration && this.currentTime ? this.currentTime >= this.duration - 1 && !this.paused : false;
```
Yuvalke: -1  - is for the buffer size where the liveEdge could be on?

RoyBregman: It should be almost the same. there is a small difference because we reduce 0.1 from the currentTime if it greater or equal to duration

Yuvalke: So I think it should be on meaningful const

src/player.js
```diff
@@ -405,6 +405,12 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _hasUserInteracted: boolean = false;
+  /**
+   * Whether the video is seeked to live edge
+   * @type {boolean}
+   * @private
+   */
+  _isOnLiveEdge: boolean = false;
```
yairans: reset it to `false` on `reset`

src/player.js
```diff
@@ -1950,6 +1972,9 @@ export default class Player extends FakeEventTarget {
       this._engine
         .load(startTime)
         .then(data => {
+          if (this.isLive() && startTime === -1) {
```
yairans:
```suggestion
          if (this.isLive() && (startTime === -1 || startTime > duration)) {
```

src/player.js

src/player.js

PR #430, opened 2019-12-31
fix(FEC-9583): race condition on load ios playback

Yuvalke:

Issue: promise resolve on the next tick so video load and then we are setting the src, no src on loading.
Solution: handle the load when promise resolve

Solve: FEC-9583.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -297,9 +297,9 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         }
         if (this._sourceObj && this._sourceObj.url) {
           this._setSrc().then(() => {
+            this._trigger(CustomEventType.ABR_MODE_CHANGED, {mode: this._isProgressivePlayback() ? 'manual' : 'auto'});
             this._videoElement.load();
           });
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -296,10 +296,13 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           this._setProgressiveSource();
         }
         if (this._sourceObj && this._sourceObj.url) {
-          this._setSrc();
+          this._setSrc().then(() => {
+            this._videoElement.load();
+          });
           this._trigger(CustomEventType.ABR_MODE_CHANGED, {mode: this._isProgressivePlayback() ? 'manual' : 'auto'});
```
OrenMe: to keep it consistent we need to also move the event to the promise.
Please check that code flow and event order hasn't changed as well - both internal handling and analytics.
Also, as `_setSrc`is called from the select progressive source, then please check that this is also working.
@yairans please review.

Yuvalke: What do you mean by checking the event order? compare to previous version?

OrenMe: Yes, please check that this async flow didn't change anything - low reason to think it did, but need to verify this.

Yuvalke: I verified the events, i didn't find different

PR #433, opened 2020-01-07
fix(FEC-9603): load doesn't call on play after attach detach

Yuvalke:

Issue: load doesn't call on play after attach detach.
Solution: add load to attachMedia.
pause and hideTextTracks added to avoid textTracks and update state of player after detach.

Solve FEC-9603.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -1016,6 +1016,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   _onCueChange(e: FakeEvent): void {
     let textTrack: TextTrack = e.currentTarget;
     let activeCues: Array<Cue> = [];
+    if (!textTrack || !textTrack.activeCues) return;
     for (let cue of textTrack.activeCues) {
       //Normalize cues to be of type of VTT model
       if (window.VTTCue && cue instanceof window.VTTCue) {
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -1016,7 +1016,6 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   _onCueChange(e: FakeEvent): void {
     let textTrack: TextTrack = e.currentTarget;
     let activeCues: Array<Cue> = [];
-    if (!textTrack || !textTrack.activeCues) return;
     for (let cue of textTrack.activeCues) {
       //Normalize cues to be of type of VTT model
       if (window.VTTCue && cue instanceof window.VTTCue) {
--- a/src/player.js
+++ b/src/player.js
@@ -547,7 +547,6 @@ export default class Player extends FakeEventTarget {
       if (!this.src) {
         this._prepareVideoElement();
       }
-      this.load();
     }
     if (this._engine) {
       this._playbackMiddleware.play(() => this._play());
@@ -2003,6 +2002,9 @@ export default class Player extends FakeEventTarget {
    * @returns {void}
    */
   _play(): void {
+    if (!this._engine.src) {
+      this.load();
+    }
     this.ready()
       .then(() => {
         if (this.isLive() && !this.isDvr()) {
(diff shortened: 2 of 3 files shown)


src/player.js
```diff
@@ -2003,6 +2002,9 @@ export default class Player extends FakeEventTarget {
    * @returns {void}
    */
   _play(): void {
+    if (!this._engine.src) {
+      this.load();
```
OrenMe: This was removed in https://github.com/kaltura/playkit-js/pull/431, so need to find a solution that will accommodate both issues.
It is not possible to return this here due to other issues it creates.

src/player.js
```diff
@@ -547,7 +547,6 @@ export default class Player extends FakeEventTarget {
       if (!this.src) {
         this._prepareVideoElement();
       }
-      this.load();
```
OrenMe: This exposes another issue though, if someone calls load and then play we will call load again in play.
@yairans FYI

src/player.js
Yuvalke: Why not? we called the internal load on the previous fix. 
this._load and then it wasn't pass through the load middleware. 
I called the load which pass through the load middleware.

OrenMe: Please see related internal ticket for details.
If you are reverting a fix that was done just a few days before you need to check why it was done.

src/player.js
Yuvalke: Right, on my case I protect it with engine.src which will be true when _loadPromise exist otherwise we'll call load.

src/player.js
Yuvalke: maybe I miss something but I saw the ticket and the issue was when calling play() we called _load internally and it wasn't used in middleware

yairans: @Yuvalke Your change makes the load middleware run after the play middleware finished. doesn't make sense.

Yuvalke: @yairans It's not after, as part of play middleware I use load middleware for specific use case.

yairans: @Yuvalke `_play` is the last link in the `play` middleware chain.
Consider bumper for example - the `play`  middleware will run before the `load`. this is unacceptable.

src/player.js
yairans: `load` is a public api so each middleware must protect itself from multiple calls as we do in our `_load` by `this_src` existence

src/player.js
yairans: can't you just call to `_load` in attach flow?

src/engines/html5/html5.js
```diff
@@ -1016,6 +1016,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
   _onCueChange(e: FakeEvent): void {
     let textTrack: TextTrack = e.currentTarget;
     let activeCues: Array<Cue> = [];
+    if (!textTrack || !textTrack.activeCues) return;
```
OrenMe: how can this happen? if you get the event it means that textTrack must exist.
If activeCues is empty we still want the event to be dispatched.

Yuvalke: Right.
activeCue is null and it's not iterable, but it should fire the evet. i'll fix it

test/src/player.spec.js
```diff
@@ -137,6 +137,21 @@ describe('Player', function() {
         player.play();
       });
     });
+    describe('attach detach', function() {
+      it('should success play after detach attach', done => {
+        const playing = () => {
+          player.removeEventListener('playing', playing);
```
OrenMe: use player event enums.

src/engines/html5/html5.js
OrenMe: active cue is not null, it is an empty array, unless the text track is disabled and then it is null but then you shouldn't get the event.

Yuvalke: it happened multiple times for me.
it occurs as part of detach sequence, it's not always happened it's depend on timing of detach request.

OrenMe: So this seems like a regression from the current change, please check.

Yuvalke: I have checked it multiple time and I can't reproduce.
I guess depend on the timing of detach and cues.

PR #435, opened 2020-01-09
fix(FEC-9609): after postroll there isn't replay button using attach detach api

Yuvalke:
#Description of the Changes

Issue: load occur on attachMediaSource so after ended event on video tag attach call load then replay missing.
Solution: add internal load to play when attach called.

Solve: FEC-9609.

### CheckLists

- [ ] Docs have been updated


src/middleware/attach-middleware.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  /**
+   * Play middleware handler.
+   * @param {Function} next - The next play handler in the middleware chain.
+   * @returns {void}
+   */
+  play(next: Function): void {
+    if (!this._context.src) {
+      this._context._load();
```
yairans: This change actually re-open FEC-9459 issue (play bypasses the `load` middleware mechanism).
Consider plugin that has `load` middleware that postpones the player loading by timeout.
By this, the loading will occur immediately.

src/middleware/attach-middleware.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+import BaseMiddleware from './base-middleware';
+
+/**
+ * Middleware implementation for bumper plugin.
```
OrenMe:
```suggestion
 * Middleware implementation for attach handler.
```

PR #439, opened 2020-01-27
fix(FEC-9629): Player is not inline when rendered after the DOM loads

shalommeoded:

This reverts commit 1b4a3ca as is unneeded any more due to https://github.com/kaltura/kaltura-player-js/pull/302

Solves FEC-9629

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -433,7 +433,8 @@ export default class Player extends FakeEventTarget {
     this._setConfigLogLevel(config);
     this._playerId = Utils.Generator.uniqueId(5);
     this._prepareVideoElement();
-    const playsInline = Utils.Object.getPropertyPath(config, 'playback.playsinline');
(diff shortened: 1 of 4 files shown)


src/player.js
```diff
@@ -433,7 +433,8 @@ export default class Player extends FakeEventTarget {
     this._setConfigLogLevel(config);
     this._playerId = Utils.Generator.uniqueId(5);
     this._prepareVideoElement();
-    const playsInline = Utils.Object.getPropertyPath(config, 'playback.playsinline');
+    const defaultPlaysInline = Utils.Object.getPropertyPath(Player._defaultConfig, 'playback.playsinline');
```
yairans: if we have an issue with our default we have to fix it for all default values not for `playsinline` only

PR #440, opened 2020-01-28
fix(FEC-9534): changing the bitrate post completion of any content changes replay to play icon

RoyBregman:

to fix bug in hls that seek event is fired when changing qualities - if state was idle (i.e end of playback) and seeked event fired don't change state to paused if current time is at the end of the playback

solves FEC-9534

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -634,6 +634,7 @@ export default class Player extends FakeEventTarget {
     this._resetStateFlags();
     this._engineType = '';
     this._streamType = '';
+    this._pendingSelectedVideoTrack = null;
     if (this._engine) {
       this._engine.reset();
     }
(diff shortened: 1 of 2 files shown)


src/player.js
```diff
@@ -1701,6 +1720,10 @@ export default class Player extends FakeEventTarget {
         if (browser === 'Edge' || browser === 'IE') {
           this._removeTextCuePatch();
         }
+        if (this._engine && this._pendingSelectedVideoTrack) {
+          this._engine.selectVideoTrack(this._pendingSelectedVideoTrack);
```
yairans: clearer to do this on `PLAY` or `PLAYING` no?

src/player.js
```diff
@@ -1130,7 +1145,11 @@ export default class Player extends FakeEventTarget {
   selectTrack(track: ?Track): void {
     if (this._engine) {
       if (track instanceof VideoTrack) {
-        this._engine.selectVideoTrack(track);
+        if (this._playbackEnded) {
+          this._pendingSelectedVideoTrack = track;
```
yairans: reset on reset

src/player.js

src/player.js

PR #448, opened 2020-03-23
fix(FEC-9605): load middleware called twice when preload is 'auto'

yairans:

don't call `load` if `src` already exists

Sloves FEC-9605

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -544,7 +544,11 @@ export default class Player extends FakeEventTarget {
         this._eventManager.listenOnce(this, CustomEventType.SOURCE_SELECTED, () => this._load());
       }
     };
-    this._playbackMiddleware.load(() => loadPlayer());
+    if (!this.src) {
+      this._playbackMiddleware.load(() => loadPlayer());
+    } else {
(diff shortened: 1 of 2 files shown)


src/player.js
```diff
@@ -558,8 +558,8 @@ export default class Player extends FakeEventTarget {
       this.dispatchEvent(new FakeEvent(CustomEventType.PLAYBACK_START));
       if (!this.src) {
         this._prepareVideoElement();
+        this.load();
```
RoyBregman: maybe it is better to put the check inside load function? as this can also be called from external api

src/player.js
```diff
@@ -544,7 +544,11 @@ export default class Player extends FakeEventTarget {
         this._eventManager.listenOnce(this, CustomEventType.SOURCE_SELECTED, () => this._load());
       }
     };
-    this._playbackMiddleware.load(() => loadPlayer());
+    if (!this.src) {
```
OrenMe: can we add a test for this?

PR #451, opened 2020-04-23
fix(FEC-9997): out of DVR window will lead to live edge

Yuvalke:

if the current time is negative then it means that the player is out of the DVR window.
DVR out of the window should seek to live edge.
Solve FEC-9997.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -2036,7 +2036,7 @@ export default class Player extends FakeEventTarget {
     }
     this.ready()
       .then(() => {
-        if (this.isLive() && !this.isDvr()) {
+        if (this.isLive() && (!this.isDvr() || (this.isDvr() && this.currentTime <= 0))) {
           this.seekToLiveEdge();
         }
         this._engine.play();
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -2036,7 +2036,7 @@ export default class Player extends FakeEventTarget {
     }
     this.ready()
       .then(() => {
-        if (this.isLive() && !this.isDvr()) {
+        if (this.isLive() && (!this.isDvr() || (this.isDvr() && typeof this.currentTime === 'number' && this.currentTime <= 0))) {
```
yairans, line 2039: `currentTime < 0` is invalid. the getter should return NaN
(also `this.isDvr()` here is redundant)

Yuvalke: current time could be negative cause we calculating it by currentTime of video Tag - getStartTimeOfDvrWindow() which means that if we have a DVR window of 2 hours and I'll seek to start and pause the DVR window is moving and current time become negative.
`isDvr` is important here cause There are two scenarios:
1. live without DVR which should always go to live edge(old one)
2. live DVR with negative currentTime(out of DVR window) (new one)

Yuvalke: You are correct `currentTime < 0` is invalid and it'll happen only on live DVR which will resolve here.

PR #452, opened 2020-04-30
fix(FEC-10014): Content playback fails when user minimise the browser on IOS

shalommeoded:

The player is broken when switching between tabs or just minimising Safari.
The original issue has occurred here FEC-8873  however was resolved only for ad use case.
In addition, the main issue is still occuring and the bug is still open with apple: https://bugs.webkit.org/show_bug.cgi?id=195452

As a current solutions I have implemented a nudge to the seek in order to bypass the issue when returning from "Idle" mode.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -307,7 +307,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         // Sometimes when playing live in safari and switching between tabs the currentTime goes back with no seek events
         this._eventManager.listen(window, 'focus', () =>
           setTimeout(() => {
-            this._videoElement.currentTime = Math.max(0, this._videoElement.currentTime - 0.1);
+            this._videoElement.currentTime = this._videoElement.currentTime > 0.1 ? this._videoElement.currentTime - 0.1 : 0;
             this._syncCurrentTime();
           }, BACK_TO_FOCUS_TIMEOUT)
         );
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -17,6 +17,7 @@ import type {FairplayDrmConfigType} from './fairplay-drm-handler';
 
 const BACK_TO_FOCUS_TIMEOUT: number = 1000;
 const MAX_MEDIA_RECOVERY_ATTEMPTS: number = 3;
+const NUDGE_SEEK_AFTER_FOCUS: number = 0.1;
 
 /**
  * An illustration of media source extension for progressive download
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -461,6 +462,10 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
   }
 
   _syncCurrentTime(): void {
+    if (SEEK_AFTER_FOCUS) {
+      this._videoElement.currentTime -= 0.1;
```
OrenMe, line 466: can't we do this in the `_handleDecodeError` when we set back current time?

shalommeoded: @OrenMe 
I tried, as it is more logical to put it there, however, the weird part of this issue, that the player never detect this issue as a decode error, or any error in fact, the player is just getting stuck with no capability of pausing or playing.

image

As you can see in the above pic, for some reason we never enter `_handleDecodeError` if we did we would have [log](https://github.com/kaltura/playkit-js/blob/84b54a1023516f3595970990dc72cdb2eabce7cb/src/engines/html5/media-source/adapters/native-adapter.js#L332) it.

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -305,7 +305,12 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         this._eventManager.listen(this._videoElement, Html5EventType.ABORT, () => this._clearHeartbeatTimeout());
         this._eventManager.listen(this._videoElement, Html5EventType.SEEKED, () => this._syncCurrentTime());
         // Sometimes when playing live in safari and switching between tabs the currentTime goes back with no seek events
-        this._eventManager.listen(window, 'focus', () => setTimeout(() => this._syncCurrentTime(), BACK_TO_FOCUS_TIMEOUT));
+        this._eventManager.listen(window, 'focus', () =>
+          setTimeout(() => {
+            this._videoElement.currentTime = Math.max(0, this._videoElement.currentTime - 0.1);
```
OrenMe, line 310: when can this happen? when currentTime is 0?

shalommeoded: Yes, just wanted to have a protection in case of an edge case, it was suggested by Roy B.

OrenMe: Ok, so you can donit with simple tenrary currenTime > 0.1 ? currentTime-0.1 : 0
What happens with live? Will this still be correct?

shalommeoded: Done,
And it is working with Live entries as well.

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -305,7 +305,12 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         this._eventManager.listen(this._videoElement, Html5EventType.ABORT, () => this._clearHeartbeatTimeout());
         this._eventManager.listen(this._videoElement, Html5EventType.SEEKED, () => this._syncCurrentTime());
         // Sometimes when playing live in safari and switching between tabs the currentTime goes back with no seek events
-        this._eventManager.listen(window, 'focus', () => setTimeout(() => this._syncCurrentTime(), BACK_TO_FOCUS_TIMEOUT));
+        this._eventManager.listen(window, 'focus', () =>
+          setTimeout(() => {
+            this._videoElement.currentTime = this._videoElement.currentTime > 0.1 ? this._videoElement.currentTime - 0.1 : 0;
```
OrenMe, line 310: one last thing is, we try to avoid fudge numbers inline, please move to const in top of file(see convention) 
`BACK_TO_FOCUS_NUDGE`

RoyBregman: Maybe also add a one liner comment just above the current time assignment to explain this bizarre behaviour we are doing - something like - "In IOS when coming back from lock screen hls sometimes gets stuck until seeking" ?

PR #455, opened 2020-05-24
fix(FEC-10122): clear drop frame watcher and DRM selection

Yuvalke:

destroy dropped frames watcher on reset to avoid a memory leak.
clear the DRM selection for the use case of loading another media without DRM keeps selection.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -211,13 +211,13 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   reset(): void {
     this._eventManager.removeAll();
-    if (this._mediaSourceAdapter) {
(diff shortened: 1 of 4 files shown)


src/engines/html5/html5.js
```diff
@@ -242,6 +243,10 @@ export default class Html5 extends FakeEventTarget implements IEngine {
       this._mediaSourceAdapter.destroy();
       this._mediaSourceAdapter = null;
     }
+    if (this._droppedFramesWatcher) {
```
OrenMe, line 246: any reason why you moved it down here? as it is dependent on _mediaSourceAdapter it could trow error if we destroy it after we destroy adapter
Same for reset above

Yuvalke: It doesn't matter but I have changed it to keep the flow of code meaningful

PR #459, opened 2020-06-23
feat(FEC-9649): loading external CSS

RoyBregman:

if provided a url in the config externalCSS we load and append the style to the document header

solves FEC-9649

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/utils/util.js
+++ b/src/utils/util.js
@@ -418,7 +418,7 @@ const _Dom = {
    * @return {Promise} - The loading promise.
    * @public
    */
-  loadStyleSheetAsync(url: string): Promise<*> {
+  loadStyleSheetAsync(url: string): Promise<HTMLLinkElement> {
     return new Promise((resolve, reject) => {
       let resolved = false,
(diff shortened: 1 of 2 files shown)


src/utils/util.js
```diff
@@ -412,6 +412,33 @@ const _Dom = {
...
+        t = document.getElementsByTagName('link')[0],
+        s = this.createElement('link');
+      s.type = 'text/css';
+      s.rel = 'stylesheet';
+      s.href = url;
+      s.async = true;
+      s.onload = s.onreadystatechange = function() {
+        if (!r && (!this.readyState || this.readyState === 'complete')) {
+          r = true;
+          resolve(this);
```
OrenMe, line 433: why are you passing `this`?

RoyBregman: It gives a ref to the link element that was added to the DOM. Actually I don't think of a use case that someone will need the reference for that... I can remove it

OrenMe: so you can set the promise type to return <null>

PR #460, opened 2020-06-28
fix: remove listeners which doesn't clear correctly

Yuvalke:

Didn't finished yet

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/engine-decorator.js
+++ b/src/engines/engine-decorator.js
@@ -22,6 +22,8 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
   constructor(engine: IEngine, pluginWithDecorators: Array<IEngineDecoratorProvider>) {
     super();
     this._eventManager = new EventManager();
+    this._eventManager.listen(engine, EventType.PLAYER_DESTROY, () => this._destroy());
+    this._eventManager.listen(engine, EventType.PLAYER_RESET, () => this._reset());
     this._pluginDecorators = pluginWithDecorators.map(plugin => plugin.getEngineDecorator(engine, super.dispatchEvent.bind(this)));
     const events: Array<string> = (Object.values(EventType): any);
     events.forEach(event => this._eventManager.listen(engine, event, e => this.dispatchEvent(e)));
@@ -46,6 +48,15 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
   dispatchEvent(event: FakeEvent): boolean {
     const activeDecorator = this._pluginDecorators.find(decorator => decorator.active);
     return activeDecorator ? activeDecorator.dispatchEvent && activeDecorator.dispatchEvent(event) : super.dispatchEvent(event);
+  }
+
+  _reset(): void {
+    this._eventManager.removeAll();
+  }
+
+  _destroy(): void {
+    this._pluginDecorators = [];
+    this._eventManager.destroy();
   }
 
   get active(): boolean {
--- a/src/engines/engine-decorator.js
+++ b/src/engines/engine-decorator.js
@@ -23,7 +23,6 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
     super();
     this._eventManager = new EventManager();
     this._eventManager.listen(engine, EventType.PLAYER_DESTROY, () => this._destroy());
-    this._eventManager.listen(engine, EventType.PLAYER_RESET, () => this._reset());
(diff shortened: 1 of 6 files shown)


src/plugin/base-plugin.js
```diff
@@ -136,7 +138,9 @@ export default class BasePlugin implements IPlugin {
    * @virtual
    * @returns {void}
    */
-  reset(): void {}
+  reset(): void {
+    this.eventManager.removeAll();
```
OrenMe, line 142: it is up to plugin to decide this

src/engines/engine-decorator.js
```diff
@@ -48,6 +50,15 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
     return activeDecorator ? activeDecorator.dispatchEvent && activeDecorator.dispatchEvent(event) : super.dispatchEvent(event);
   }
 
+  _reset(): void {
+    this._eventManager.removeAll();
```
OrenMe, line 54: same here, do we event reset the proxy and re-register to the events? 
Also, the dependency is backward here, player creates this instance so need to call from player on destroy of proxy and not listen to events.

src/engines/html5/media-source/base-media-source-adapter.js
```diff
@@ -186,11 +187,13 @@ export default class BaseMediaSourceAdapter extends FakeEventTarget implements I
    * @private
    */
   _handleLiveTimeUpdate(): void {
-    this._videoElement.addEventListener(Html5EventType.DURATION_CHANGE, () => {
-      if (this.isLive() && this._videoElement.paused) {
-        this._trigger(Html5EventType.TIME_UPDATE);
-      }
-    });
+    this._videoElement.addEventListener(Html5EventType.DURATION_CHANGE, this._onDurationChanged.bind(this));
```
RoyBregman, line 190: Why not use event manager also here?

src/engines/html5/media-source/adapters/fairplay-drm-handler.js
```diff
@@ -114,7 +117,7 @@ class FairplayDrmHandler {
     let message = event.message;
     let request = new XMLHttpRequest();
     request.responseType = 'arraybuffer';
-    request.addEventListener('load', (e: Event) => this._licenseRequestLoaded(e), false);
+    request.addEventListener('load', (e: Event) => this._licenseRequestLoaded(e));
```
RoyBregman, line 120: Why not use event manager also here?

RoyBregman, line 120: Why not use event manager also here?

src/engines/html5/media-source/base-media-source-adapter.js
Yuvalke: If media source adapter will do remove all listener this listener will be deleted and nobody will re-register it. 
I think we want to make sure it'll not happened

src/engines/html5/media-source/adapters/fairplay-drm-handler.js
Yuvalke: Could be many requests and after it's finish the request object is not relevant anymore, it's a local var.
It'll fill the event manager with non relevant handlers

RoyBregman: I don't understand, like the way it is now the listener will be never removed...

Yuvalke: Right, it should be removed.

src/engines/engine-decorator.js
```diff
@@ -22,6 +22,7 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
   constructor(engine: IEngine, pluginWithDecorators: Array<IEngineDecoratorProvider>) {
     super();
     this._eventManager = new EventManager();
+    this._eventManager.listen(engine, EventType.PLAYER_DESTROY, () => this._destroy());
```
OrenMe, line 25: @Yuvalke why listen to the underlying engine when this is explicitly called by the player?
Add this as a method and call the underlying engine explicitly.

src/engines/html5/media-source/base-media-source-adapter.js
OrenMe: this won't remove the listener, you need to keep reference to the bounded method and use that in the add and remove

PR #465, opened 2020-07-16
fix(FEC-10312): ready promise rejected in recoverable error

yairans:

The ready promise rejected only when a critical error occurred
this makes the player ready before it actually loads the source

Solves FEC-10312

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/test/src/player.spec.js
+++ b/test/src/player.spec.js
@@ -209,6 +209,24 @@ describe('Player', function() {
               done();
             });
           });
+
+          it('should resolved successfully even recoverable error occurred', done => {
+            let readyResolved = false;
+            player.addEventListener(CustomEventType.TRACKS_CHANGED, () => {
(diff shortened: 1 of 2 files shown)


OrenMe:
good catch @yairans, in which use cases can this happen?
Please open internal tracking ticket as well.

yairans:
@OrenMe I caught it in  unregistered -plugin error

test/src/player.spec.js
```diff
@@ -209,6 +209,24 @@ describe('Player', function() {
               done();
             });
           });
+
+          it('should resolved successfully even recoverable error occurred', done => {
```
OrenMe, line 213:
```suggestion
          it('should resolved successfully even when recoverable error occurred', done => {
```

OrenMe:
Flow failing :-(

PR #469, opened 2020-07-21
fix(FEC-10267): Ad controls are disappearing in DAI LIVE playback

Yuvalke:

Issue: IMA DAI live, player seek to live edge during ad since it's our expected behavior on live.
Solution: add a check for an active ad break, don't seek an active ad break.

Solve FEC-10267

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -2073,8 +2073,13 @@ export default class Player extends FakeEventTarget {
     this.ready()
       .then(() => {
         const liveOrDvrOutOfDvrWindow = this.isLive() && (!this.isDvr() || (typeof this.currentTime === 'number' && this.currentTime < 0));
-        if (!(this._adsController && this._adsController.isAdBreak()) && liveOrDvrOutOfDvrWindow) {
-          this.seekToLiveEdge();
+        if (liveOrDvrOutOfDvrWindow) {
+          if (this._adsController && this._adsController.isAdBreak()) {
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -2072,7 +2072,8 @@ export default class Player extends FakeEventTarget {
     }
     this.ready()
       .then(() => {
-        if (this.isLive() && (!this.isDvr() || (typeof this.currentTime === 'number' && this.currentTime < 0))) {
+        const liveOrDvrOutOfDvrWindow = (this.isLive() && (!this.isDvr() || (typeof this.currentTime === 'number' && this.currentTime < 0)));
+        if (!this._adsController.isAdBreak() && liveOrDvrOutOfDvrWindow) {
           this.seekToLiveEdge();
```
yairans, line 2077: should seek once the ad is done no?

Yuvalke: I don't think so I think DAI keeps you on the edge when you're playing ads.

yairans: what is the user paused for more than the non-dvr window (the actuall buffer length)
this change will try to play out of buffer content

yairans: and need to test on safari as well as it plays natively

Yuvalke: Fix by after ad break end I'm seeking to live edge - tested on safari as well

OrenMe:
this behaviour is edge case for IMA DAI, we will defer this handing for DAI solution as this solution doesn't behave correctly with live without DVR

PR #471, opened 2020-07-22
fix(FEC-10335): ad doesn't play correctly on same video tag with MSE

Yuvalke:

IMA probably changed their adloaded timing which causes issues on our logic of playing ads with MSE.
issue: video detached 6-8 sec before midroll supposes to show up which causes an empty video tag.
Solution: detach the media when the ad break starts.

Solve FEC-10335.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1811,11 +1811,10 @@ export default class Player extends FakeEventTarget {
         }
       });
       if (this.config.playback.playAdsWithMSE) {
-        this._eventManager.listen(this, AdEventType.AD_LOADED, event => {
-          if (event.payload.ad.linear) {
-            this._eventManager.listenOnce(this, AdEventType.AD_BREAK_START, () => {
-              this._detachMediaSource();
-            });
+        this._eventManager.listen(this, AdEventType.AD_BREAK_START, () => {
+          const adData = this._adsController ? this._adsController.getAd() : null;
+          if (!!adData && adData.linear) {
+            this._detachMediaSource();
           }
         });
         this._eventManager.listen(this, AdEventType.AD_BREAK_END, () => this._attachMediaSource());
--- a/src/player.js
+++ b/src/player.js
@@ -1811,13 +1811,15 @@ export default class Player extends FakeEventTarget {
         }
       });
       if (this.config.playback.playAdsWithMSE) {
-        this._eventManager.listen(this, AdEventType.AD_LOADED, (event: FakeEvent) => {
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -1811,13 +1811,15 @@ export default class Player extends FakeEventTarget {
         }
       });
       if (this.config.playback.playAdsWithMSE) {
-        this._eventManager.listen(this, AdEventType.AD_LOADED, (event: FakeEvent) => {
+        this._eventManager.listen(this, AdEventType.AD_LOADED, event => {
           if (event.payload.ad.linear) {
-            this._detachMediaSource();
+            this._eventManager.listenOnce(this, AdEventType.AD_BREAK_START, () => {
+              this._detachMediaSource();
+            });
```
OrenMe, line 1818: do we still need the AD_LOADED? why not register to AD_BREAK_START only?
If it is because of the ad.linear that only comes in AD_LOADED then I assume we can take the ad type from the current ad we have in the ad controller.

src/player.js
```diff
@@ -1811,13 +1811,15 @@ export default class Player extends FakeEventTarget {
         }
       });
       if (this.config.playback.playAdsWithMSE) {
-        this._eventManager.listen(this, AdEventType.AD_LOADED, (event: FakeEvent) => {
+        this._eventManager.listen(this, AdEventType.AD_LOADED, event => {
```
OrenMe, line 1814: why did you remove the type?

src/player.js
```diff
@@ -1811,13 +1811,15 @@ export default class Player extends FakeEventTarget {
...
           if (event.payload.ad.linear) {
-            this._detachMediaSource();
+            this._eventManager.listenOnce(this, AdEventType.AD_BREAK_START, () => {
+              this._detachMediaSource();
+            });
           }
         });
-        this._eventManager.listen(this, AdEventType.AD_BREAK_END, this._attachMediaSource);
-        this._eventManager.listen(this, AdEventType.AD_ERROR, this._attachMediaSource);
+        this._eventManager.listen(this, AdEventType.AD_BREAK_END, () => this._attachMediaSource());
```
OrenMe, line 1821: was this a mistake? the target is this so the context should remain the same in the handler

src/player.js
Yuvalke: Mistake, I'll add it

src/player.js
Yuvalke: both correct wanted to be more explicit.

PR #474, opened 2020-08-03
fix(FEC-10275): Bumper incorrectly recognised as ad 

dan-ziv:

Player's `getAd()` API shouldn't return `Ad` object while ad is only loaded, this can cause issues when IMA loads its ads so early:
1. `AD_LOADED` is triggered while media is playing - `getAd()` returns ad but no ad is currently playing.
2. `AD_LOADED` is triggered while another ad/bumper is playing - `getAd()` returns the wrong ad.

Solution:
* To access the loaded ad from the player's API, use `getPendingAd()` API. In such way, we are solving the above issues:
1. `AD_LOADED` is triggered while media is playing - `getAd()` returns nothing, `getPendingAd()` returns the loaded ad.
2. `AD_LOADED` is triggered while another ad/bumper is playing - `getAd()` returns the current ad, `getPendingAd()` returns the next ad.

* change `playAdsWithMSE` implementation to listen to `AD_LOADED` event and only if the pending ad is linear, listen to ad break start to run the `playAdsWithMSE` logic.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/ads/ads-controller.js
+++ b/src/ads/ads-controller.js
@@ -29,6 +29,7 @@ class AdsController extends FakeEventTarget implements IAdsController {
   _adBreaksLayout: Array<number | string>;
   _adBreak: ?AdBreak;
   _ad: ?Ad;
+  _pendingAd: ?Ad;
   _adPlayed: boolean;
   _snapback: number;
   _configAdBreaks: Array<RunTimeAdBreakObject>;
@@ -85,6 +86,15 @@ class AdsController extends FakeEventTarget implements IAdsController {
    */
   getAd(): ?Ad {
     return this._ad;
+  }
+
+  /**
+   * @instance
+   * @memberof AdsController
+   * @returns {?Ad} - Gets the pending ad data.
+   */
+  getPendingAd(): ?Ad {
+    return this._pendingAd;
   }
 
   /**
@@ -285,11 +295,12 @@ class AdsController extends FakeEventTarget implements IAdsController {
 
   _onAdLoaded(event: FakeEvent): void {
     this._adIsLoading = false;
-    this._ad = event.payload.ad;
+    this._pendingAd = event.payload.ad;
   }
 
   _onAdStarted(event: FakeEvent): void {
     this._ad = event.payload.ad;
+    this._pendingAd = null;
     this._adPlayed = true;
   }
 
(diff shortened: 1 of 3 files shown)


src/ads/ads-controller.js
```diff
@@ -285,11 +295,12 @@ class AdsController extends FakeEventTarget implements IAdsController {
 
   _onAdLoaded(event: FakeEvent): void {
     this._adIsLoading = false;
-    this._ad = event.payload.ad;
+    this._pendingAd = event.payload.ad;
   }
 
   _onAdStarted(event: FakeEvent): void {
     this._ad = event.payload.ad;
+    this._pendingAd = null;
```
Yuvalke, line 303: You need to check if the pending ad is started(could be more than 1 pending ad), it could happen when multiple ad providers exist like bumper and ima.

src/player.js
```diff
@@ -1811,10 +1811,12 @@ export default class Player extends FakeEventTarget {
         }
       });
       if (this.config.playback.playAdsWithMSE) {
-        this._eventManager.listen(this, AdEventType.AD_BREAK_START, () => {
-          const adData = this._adsController ? this._adsController.getAd() : null;
-          if (!!adData && adData.linear) {
-            this._detachMediaSource();
+        this._eventManager.listen(this, AdEventType.AD_LOADED, (event: FakeEvent) => {
```
Yuvalke, line 1814: as I mention, multiple ads could be loaded but you want to activate the listener to the fit one. 
I think it'll be better to keep the pending ads as sorted array and take the first one in ad_break_start.

src/ads/ads-controller.js
dan-ziv: That's true that u can have multiple providers, but at a certain point of time just one ad can be loaded. I don't think we need to complicate it too much.

Yuvalke: This issue exists, for example, bumper with ima both postroll which one will be in pendingAd?
if bumper is loading after ima postroll, ima won't be exist in pendingAd and then the logic for smart won't work for postroll.

src/player.js
yairans: 1. this code move to kaltura player
2. `AD_LOADED` listening was the previous implement before @Yuvalke changed it to `AD_BREAK_START` in https://github.com/kaltura/playkit-js/pull/471 to fix a bug

yairans (changes requested):

PR #478, opened 2020-08-11
feat(FEC-10146): Add native video track analytics data

Yuvalke:

issue: the native player doesn't provide analytics data about the video track.
solution: check the videoHeight, videoWidth if it changed the track change event triggered.

Solve: FEC-10146.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -502,7 +502,14 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     const {videoHeight, videoWidth} = this._videoElement;
     if (!this._videoDimensions || videoHeight !== this._videoDimensions.videoHeight || videoWidth !== this._videoDimensions.videoWidth) {
       this._videoDimensions = {videoHeight, videoWidth};
-      this._onTrackChanged(new VideoTrack({height: videoHeight, width: videoWidth, active: true}));
+      let setting = {
+        language: '',
+        height: videoHeight,
+        width: videoWidth,
+        active: true
+      };
+      this._onTrackChanged(new VideoTrack(setting));
+      NativeAdapter._logger.debug('Video track change', new VideoTrack(setting));
     }
   }
 
@@ -597,11 +604,13 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     if (videoTracks) {
       for (let i = 0; i < videoTracks.length; i++) {
         const settings = {
-          //TODO calculate width/height/bandwidth
+          //TODO calculate bandwidth
           id: videoTracks[i].id,
           active: videoTracks[i].selected,
           label: videoTracks[i].label,
           language: videoTracks[i].language,
+          width: videoTracks[i].width | this._videoElement.videoWidth,
+          height: videoTracks[i].height | this._videoElement.videoHeight,
           index: i
         };
         parsedTracks.push(new VideoTrack(settings));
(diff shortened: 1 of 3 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -460,6 +466,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         this._trigger(Html5EventType.WAITING);
       }
     }
+    this._handleVideoTracksChange();
```
OrenMe, line 469: Use track events instead

src/player.js
```diff
@@ -2132,18 +2132,25 @@ export default class Player extends FakeEventTarget {
...
       for (let i = 0; i < tracks.length; i++) {
-        tracks[i].active = track.index === tracks[i].index;
+        tracks[i].active = track.index >= 0 ? track.index === tracks[i].index : track.label === tracks[i].label;
+        markedActive = markedActive ? markedActive : tracks[i].active;
       }
     }
+    // selected track for some cases won't be available on the parsed tracks, added to get them from the API.
+    if (!markedActive) {
+      this._tracks.push(track);
+    }
```
OrenMe, line 2153: this is not really needed as in native there is always one active track so this API doesn't even really make sense.
Unless I'm missing something - this is not required and the incentive for adding the track change event was only for analytics purposes where rendition changes are what maters.
We can then see for example correlation between buffering events and track changes or see excessive rendition changes which can show some bad encoding ladder config.

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -581,11 +604,13 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
       for (let i = 0; i < videoTracks.length; i++) {
         const settings = {
-          //TODO calculate width/height/bandwidth
+          //TODO calculate bandwidth
           id: videoTracks[i].id,
           active: videoTracks[i].selected,
           label: videoTracks[i].label,
           language: videoTracks[i].language,
+          width: videoTracks[i].width | this._videoElement.videoWidth,
+          height: videoTracks[i].height | this._videoElement.videoHeight,
```
OrenMe, line 613: this will only include, at best, one track so see previous comment - I think we can omit this.

OrenMe (changes requested):

src/player.js
Yuvalke: Native video tracks have one video track on the beginning after parsing, so on the change track, we don't have the track after parsing so we need to add the selected tracks to `_tracks` object for other API's. (getActiveTrack, getVideoTracks)

src/engines/html5/media-source/adapters/native-adapter.js
Yuvalke: Why not? it'll indicate which track is playing

OrenMe: see my other comment - the intention of this requirement is to have the transient event logged, and not accumulate them in the tracks object.

OrenMe (approved):

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -491,6 +498,21 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     }
   }
 
+  _handleVideoTracksChange(): void {
+    const {videoHeight, videoWidth} = this._videoElement;
+    if (!this._videoDimensions || videoHeight !== this._videoDimensions.videoHeight || videoWidth !== this._videoDimensions.videoWidth) {
+      this._videoDimensions = {videoHeight, videoWidth};
+      let setting = {
```
dan-ziv, line 505: why let and not const?

src/engines/html5/media-source/adapters/native-adapter.js
dan-ziv, line 469: Do we must check it on every time update instead of tracks changed?

Yuvalke: SmartTV doesn't have tracks change

OrenMe (approved):

PR #483, opened 2020-09-24
fix(FEC-10510): iOS14 -  cannot exit from native PiP by the player's PiP button

yairans:

* Don't use `exitPictureInPicture` when it's native PiP but use `webkitSetPresentationMode('inline')`
* Remove the redundant `LEAVE_PICTURE_IN_PICTURE` firing. it's already fired natively. 

Solves FEC-10510

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -530,12 +530,10 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   exitPictureInPicture(): void {
     try {
-      if (typeof this._el.webkitSetPresentationMode === 'function') {
-        this._el.webkitSetPresentationMode('inline');
-      } else if (document.pictureInPictureEnabled && typeof document.exitPictureInPicture === 'function') {
-        // Currently it's supported in chrome and in safari. So if we consider checking support before,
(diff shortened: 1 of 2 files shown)


src/engines/html5/html5.js
```diff
@@ -541,10 +543,6 @@ export default class Html5 extends FakeEventTarget implements IEngine {
             )
           );
         });
-      } else if (typeof this._el.webkitSetPresentationMode === 'function') {
-        this._el.webkitSetPresentationMode('inline');
-        // Safari does not fire this event but Chrome does, normalizing the behaviour
-        setTimeout(() => this.dispatchEvent(new FakeEvent(Html5EventType.LEAVE_PICTURE_IN_PICTURE)), 0);
```
OrenMe, line 547: we don't need the event anymore?

yairans: Actually, I don't see where this code was relevant, in all safari I've checked (mac, iPad with ios 11, 12, 13, and ios 14) this code was unreachable (means the `if` was applied not the `else if`). 
After this change I see this event fired natively.

Yuvalke (dismissed):

OrenMe (approved):

PR #489, opened 2020-09-30
fix(FEC-10516): Fullscreen event dispatched twice from the player

dan-ziv:

Add flag to overcome browsers which supports more than one fullscreenchange event.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/fullscreen/fullscreen-controller.js
+++ b/src/fullscreen/fullscreen-controller.js
@@ -19,6 +19,8 @@ class FullscreenController {
   _player: Player;
   _isInBrowserFullscreen: boolean;
   _eventManager: EventManager;
+  // Flag to overcome browsers which supports more than one fullscreenchange event
+  _isFullscreenEventDispatched: boolean = false;
 
   /**
    * after component mounted, set up event listeners to window fullscreen state change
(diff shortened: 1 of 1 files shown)


src/fullscreen/fullscreen-controller.js
```diff
@@ -19,6 +19,8 @@ class FullscreenController {
   _player: Player;
   _isInBrowserFullscreen: boolean;
   _eventManager: EventManager;
+  // Flag to overcome browsers which supports more than one fullscreenchange event
+  _isFullscreenEventDispatched: boolean = false;
```
Yuvalke, line 23: Which browsers does it?
It was in the past and I removed it since shaka polyfill cause this issue.

Yuvalke (changes requested):

OrenMe (approved):

src/fullscreen/fullscreen-controller.js
dan-ziv: Has nothing to do with Shaka AFAIK. 
Happens on Safari which supports `fullcreenchange` and `webkitfullscreenchange`
and on IE/Edge  which supports `fullcreenchange` and `msfullscreenchange`.

Yuvalke: Shaka fixed this polyfill on their side

OrenMe: these are double events we getting from registering them all.
We will open a task to register the correct events like screenfull library is doing.

PR #495, opened 2020-10-25
fix(FEC-10427): remove cues for native external subtitles 

Yuvalke:

Issue: safari keeps the cues from external text tracks that were active and keeps playing it on the next media.
Solution: remove cues from next media won't use the text tracks that was active.

### CheckLists

- [ ] Docs have been updated


src/player.js
```diff
@@ -574,7 +574,8 @@ export default class Player extends FakeEventTarget {
     if (this._reset) return;
     this.pause();
     //make sure all services are reset before engine and engine attributes are reset
-    this._externalCaptionsHandler.reset();
+    // $FlowFixMe
+    this._externalCaptionsHandler.reset(this._tracks.filter(track => track.external));
```
OrenMe, line 578: what is the issue with the typing?

src/track/external-captions-handler.js
```diff
@@ -456,6 +458,23 @@ class ExternalCaptionsHandler extends FakeEventTarget {
...
+   * removing cues from an existing text element in a video tag
+   * @param {Array<TextTrack>} externalTracks - external tracks
+   * @return {void}
+   */
+  _removeCuesFromExternalNativeTextTrack(externalTracks: Array<TextTrack>): void {
+    const videoElement = this._player.getVideoElement();
+    if (videoElement) {
+      Array.from(externalTracks).forEach(externalTrack => {
+        const track = Array.from(videoElement.textTracks).find(track => (track ? track.language === externalTrack.language : false));
+        if (track.cues) {
```
yairans, line 471: check `track` is not null

src/track/external-captions-handler.js
```diff
@@ -456,6 +458,23 @@ class ExternalCaptionsHandler extends FakeEventTarget {
...
+  /**
+   * removing cues from an existing text element in a video tag
+   * @param {Array<TextTrack>} externalTracks - external tracks
+   * @return {void}
+   */
+  _removeCuesFromExternalNativeTextTrack(externalTracks: Array<TextTrack>): void {
+    const videoElement = this._player.getVideoElement();
+    if (videoElement) {
+      Array.from(externalTracks).forEach(externalTrack => {
+        const track = Array.from(videoElement.textTracks).find(track => (track ? track.language === externalTrack.language : false));
```
yairans, line 470:
```suggestion
        const track = Array.from(videoElement.textTracks).find(track => (track && track.language === externalTrack.language));
```

yairans (changes requested):

src/player.js
Yuvalke: the external property doesn't exist on Track class sp flow doesn't like it but filter using `undefined` same as `false` so there is no problem.

src/track/external-captions-handler.js
```diff
@@ -456,6 +460,23 @@ class ExternalCaptionsHandler extends FakeEventTarget {
...
 
+  /**
+   * set external prefix on reset to avoid usage of the text track on the next media
+   * @param {Array<TextTrack>} externalTracks - external tracks
+   * @return {void}
+   */
+  _setPrefixForExternalNativeTextTrack(externalTracks: Array<TextTrack>): void {
+    const videoElement = this._player.getVideoElement();
+    if (videoElement) {
+      Array.from(externalTracks).forEach(externalTrack => {
```
yairans, line 471: seems like `Array.from` should be before. this method already gets `Array`

yairans (changes requested):

src/track/external-captions-handler.js
```diff
@@ -240,7 +240,7 @@ class ExternalCaptionsHandler extends FakeEventTarget {
   reset(externalTracks: Array<TextTrack>): void {
     this._resetCurrentTrack();
     this._textTrackModel = {};
-    this._setPrefixForExternalNativeTextTrack(externalTracks);
+    this._deleteExternalNativeTextTrack(externalTracks);
```
OrenMe, line 243: delete is not the best name I think as we are not deleting it not does the HTML5 spec exposes a delete method.
```suggestion
    this._resetExternalNativeTextTrack(externalTracks);
```

OrenMe (changes requested):

PR #499, opened 2020-11-01
feat(FEC-10486): Lock orientation according to screenLockOrientionMode config when switching to full screen

Yuvalke:

add screenLockOrientionMode config to be able to lock the screen mode in fullscreen.
new enum with possible values for this config:
`"none"
"any"
"natural"
"landscape"
"portrait"
"portrait-primary"
"portrait-secondary"
"landscape-primary"
"landscape-secondary"`

Solves FEC-10486

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -898,7 +898,7 @@ var config = {
 > > screenLockOrientionMode: string;
 > > ```
 > >
-> > > ##### Description: Gives the ability to lock the screen on fullscreen
+> > > ##### Description: Gives the ability to lock the screen orientation in fullscreen
 >
 > ##
 >
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1403,4 +1403,53 @@ var config = {
 
 ##
 
+> ### config.dimensions
+>
+> ##### Type: `PKDimensionsConfig`
+>
+> ```js
+> {
+>   width?: string | number;
+>   height?: string | number;
+>   ratio?: string;
+> }
+> ```
+>
+> ##### Description: Dimensions configuration
+>
+> > ### config.dimensions.width
+> >
+> > ##### Type: `string | number`
+> >
+> > ##### Default: `''`
+> >
+> > ##### Description: The width of the player.
+> >
+> > If number was provided, the width will be calculated in pixels (`width: 640` equivalent to `width: '640px'`).
+> > If string was provided, any valid css syntax can be passed, for example: `width: '100%'`, `width: 'auto'`, etc.
+> >
+> > ### config.dimensions.height
+> >
+> > ##### Type: `string | number`
+> >
+> > ##### Default: `''`
+> >
+> > ##### Description: The height of the player.
+> >
+> > If number was provided, the height will be calculated in pixels (`height: 360` equivalent to `width: '360px'`).
+> > If string was provided, any valid css syntax can be passed, for example: `height: '100%'`, `height: 'auto'`, etc.
+> >
+> > ### config.dimensions.ratio
+> >
+> > ##### Type: `string`
+> >
+> > ##### Default: `''`
+> >
+> > ##### Description: Defines the aspect ratio of the player.
+> >
+> > The aspect ratio should be written in the form of `'width:height'`, for example: `'4:3'` (classic TV ratio).
+> > If one of the `height` or `width` parameters is additionally provided in the configuration, the value of the other parameter not provided will be calculated accordingly to match the aspect ratio. If both were provided, the `height` value would be overridden.
+
+##
+
 Now that we've learned about the different options available in the player configuration, let's see [how does the source selection logic works](./source-selection-logic.md).
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -486,7 +486,8 @@ var config = {
 >  streamPriority: Array<PKStreamPriorityObject>,
 >  preferNative: PKPreferNativeConfigObject,
 >  inBrowserFullscreen: boolean,
->  playAdsWithMSE: boolean
+>  playAdsWithMSE: boolean,
+> screenLockOrientionMode: string
 > }
 > ```
 >
@@ -511,6 +512,7 @@ var config = {
 >  muted: false,
 >  pictureInPicture: true,
 >  playAdsWithMSE: false,
+> screenLockOrientionMode: '',
 >  options: {
 >    html5: {
 >      hls: {},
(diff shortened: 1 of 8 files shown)


OrenMe (commented):
I think we should change the config to accepts the orientation as a config option that accepts any of the acceptable strings as described here: https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock
```
"any"
"natural"
"landscape"
"portrait"
"portrait-primary"
"portrait-secondary"
"landscape-primary"
"landscape-secondary"
```
if it is set then use the value on the request for lock()

src/fullscreen/fullscreen-controller.js
```diff
@@ -164,7 +164,14 @@ class FullscreenController {
       this._player.exitPictureInPicture();
     }
     Promise.resolve(this._nativeEnterFullScreen(fullScreenElement)).then(
-      () => (this._isInFullscreen = true),
+      () => {
+        this._isInFullscreen = true;
+        const screenOrientation = screen && screen.orientation && typeof screen.orientation.lock === 'function';
+        const playbackConfig = this._player.config.playback;
+        if (screenOrientation && !playbackConfig.disableFullScreenOrientation) {
+          screen.orientation.lock('landscape');
```
RoyBregman, line 172: isn't it screen.lockOrientation?

src/fullscreen/fullscreen-controller.js
```diff
@@ -195,7 +202,13 @@ class FullscreenController {
    */
   _requestExitFullscreen(): void {
     Promise.resolve(this._nativeExitFullScreen()).then(
-      () => (this._isInFullscreen = false),
+      () => {
+        this._isInFullscreen = false;
+        const screenOrientation = screen && screen.orientation && typeof screen.orientation.lock === 'function';
```
RoyBregman, line 207: shouldnt you check unlock here (although if lock exists it is same)?

src/fullscreen/fullscreen-controller.js
```diff
@@ -164,7 +164,14 @@ class FullscreenController {
       this._player.exitPictureInPicture();
     }
     Promise.resolve(this._nativeEnterFullScreen(fullScreenElement)).then(
-      () => (this._isInFullscreen = true),
+      () => {
+        this._isInFullscreen = true;
+        const screenOrientation = screen && screen.orientation && typeof screen.orientation.lock === 'function';
```
OrenMe, line 169: we can do this once in the top of the module - this check value doesn't change.

src/fullscreen/fullscreen-controller.js
OrenMe, line 207: you meant `unloack`? and same as previous, can be done at top once

OrenMe: maybe it is enough to check for `screen && screen.orientation`?

src/player-config.js
```diff
@@ -37,7 +37,7 @@ const DefaultConfig = {
       dash: false
     },
     inBrowserFullscreen: false,
-    disableFullScreenOrientation: false,
+    screenLockOrientionMode: '',
```
OrenMe, line 40: actually default can be any I guess, as this is the default, right?

src/fullscreen/fullscreen-controller.js
```diff
@@ -204,8 +206,7 @@ class FullscreenController {
     Promise.resolve(this._nativeExitFullScreen()).then(
       () => {
         this._isInFullscreen = false;
-        const screenOrientation = screen && screen.orientation && typeof screen.orientation.lock === 'function';
-        if (screenOrientation) {
+        if (this._isScreenOrientationSupport) {
```
OrenMe, line 209: We should check if we locked it before unlocking as there could be a case where app locked it and we shouldn't change that.

Yuvalke: If you set any as default it would be redundant to check if I locked before we'll always make the lock...
so I changed it back to an invalid value to avoid this lock by default.

src/fullscreen/fullscreen-controller.js
Yuvalke: There isn't a good way to check it... `any` is the default that could be any value, just keeping the state like it was.
The way to check it to listen to orientation change and keep an indicator if it changed not on fullscreen call.

src/fullscreen/fullscreen-controller.js
Yuvalke: nope

RoyBregman (approved):

OrenMe (changes requested):

src/fullscreen/fullscreen-controller.js
```diff
@@ -22,6 +23,10 @@ class FullscreenController {
   // Flag to indicate that player is in fullscreen(when different element on fullscreen - api return correct state).
   _isInFullscreen: boolean = false;
   _isInBrowserFullscreen: boolean;
+  _isScreenLocked: boolean = false;
+  _isScreenOrientationSupport: boolean =
+    // $FlowFixMe
+    screen && screen.orientation && typeof screen.orientation.unlock === 'function' && typeof screen.orientation.lock === 'function';
```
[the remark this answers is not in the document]
Yuvalke: Done

docs/configuration.md
```diff
@@ -486,7 +486,8 @@ var config = {
 >  streamPriority: Array<PKStreamPriorityObject>,
 >  preferNative: PKPreferNativeConfigObject,
 >  inBrowserFullscreen: boolean,
->  playAdsWithMSE: boolean
+>  playAdsWithMSE: boolean,
+> screenLockOrientionMode: string
```
OrenMe, line 490: indent

docs/configuration.md
```diff
@@ -511,6 +512,7 @@ var config = {
 >  muted: false,
 >  pictureInPicture: true,
 >  playAdsWithMSE: false,
+> screenLockOrientionMode: '',
```
OrenMe, line 515: indent

docs/configuration.md
```diff
@@ -511,6 +512,7 @@ var config = {
 >  muted: false,
 >  pictureInPicture: true,
 >  playAdsWithMSE: false,
+>  screenLockOrientionMode: '',
```
OrenMe, line 515: Need to change to the none enum

docs/configuration.md
```diff
@@ -886,6 +888,20 @@ var config = {
...
+> >
+> > ##### Type: `string` - value list option in ScreenOrientationType
+> >
+> > ##### Default: ``
+> >
+> > ```js
+> > screenLockOrientionMode: string;
+> > ```
+> >
+> > > ##### Description: Gives the ability to lock the screen on fullscreen
```
OrenMe, line 901:
```suggestion
> > > ##### Description: Gives the ability to lock the screen orientation in fullscreen
```

OrenMe (changes requested):

docs/configuration.md
```diff
@@ -886,6 +888,20 @@ var config = {
 >
 > ##
 >
+> > ### config.playback.screenLockOrientionMode
+> >
+> > ##### Type: `string` - value list option in ScreenOrientationType
+> >
+> > ##### Default: ``
```
OrenMe, line 895: fix default

OrenMe (approved):

PR #503, opened 2020-11-10
fix(FEC-10662): source removed before destroy process finish

Yuvalke:

Issue: source has been removed from video tag before destroy has been finished.
Solution: wait for mediaSource destroy before removing the src from video tag.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -489,7 +489,6 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     this._el.load();
     return this._canLoadMediaSourceAdapterPromise
       .then(() => {
-        this._reset = false;
         if (this._mediaSourceAdapter) {
           return this._mediaSourceAdapter.load(startTime).catch(error => {
             return Promise.reject(error);
@@ -985,6 +984,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    */
   _init(source: PKMediaSourceObject, config: Object): void {
     this._config = config;
+    this._reset = false;
     this._loadMediaSourceAdapter(source);
     this.attach();
   }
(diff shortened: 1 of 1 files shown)


src/engines/html5/html5.js
```diff
@@ -208,19 +208,31 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+          },
+          () => {
+            reject();
+          }
+        );
+      } else {
+        resolve();
+      }
+    });
+    this._mediaSourceAdapter = null;
```
OrenMe, line 235: we are still not waiting for the `mediaSourceAdapterDestroyed` to be resolved then in case there is no `src`.
Not sure if it matters or not but it is not consistent in terms of flow control.

Yuvalke: why not? if I don't have src it could be handled on adapters or even 3rd party libs, bottom line I'm in a state of no src on video tag same as destroy, so for the app it's the same spot as destroyed completely.

OrenMe: More around consistency than anything else, can you explain why inconsistency here is better? does it serve a purpose?

Yuvalke: it keeps the previous logic, before if src didn't exist it didn't stop the load or anything else, just did nothing...

Yuvalke: Here if it'll reject we'll stop the load process.

OrenMe: the old logic obviously wasn't the poet's intention - the promise was always meant to be the shaka destroy promise from playkit-js-dash

Yuvalke: right, so I kept this logic...
I don't think that if we have or haven't source should influence the destroy state.

OrenMe: It has nothing to do with src.
issue is that the the `_canLoadMediaSourceAdapterPromise` should only be resolved once the media adapter promise ends.
Now we are still short circuiting it if we immediately resolve it and not wait for the adapter to resolve!

OrenMe (approved):

PR #508, opened 2020-11-22
feat(FEC-10682): signal seek in different way from buffering.

Yuvalke:

Issue: seek and the buffer was signaled as the same state in the player - couldn't be able to find the difference between the real buffer and seeking.
Solution: add a new state for seeking to make sure it won't signal as a buffer.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/state/state-manager.js
+++ b/src/state/state-manager.js
@@ -7,7 +7,6 @@ import {CustomEventType, Html5EventType} from '../event/event-type';
 import FakeEvent from '../event/fake-event';
 import getLogger from '../utils/logger';
 
-const SEEKING_THRESHOLD: number = 500;
 /**
  * This class responsible to manage all the state machine of the player.
  * @classdesc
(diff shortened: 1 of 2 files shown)


src/state/state-manager.js
```diff
@@ -111,9 +112,18 @@ export default class StateManager {
...
-        this._dispatchEvent();
+        if (this._player.seeking) {
+          this._updateState(StateType.SEEKING);
+          this._dispatchEvent();
+          //if waiting and seeking took more than threshold, it's buffering
+          setTimeout(() => {
+            if (StateType.SEEKING === this._curState.type) {
+              this._waiting();
+            }
+          }, SEEKING_THRESHOLD);
```
OrenMe, line 123: I think it is better to add spinner in UI to account for seeking then to manage by timeout

OrenMe: another option is to add reason to event,
for example `{reason: seek}` and then in UI we don't need to change a thing and in analytics we can check reason and not trigger buffer under run.

OrenMe (approved):

PR #513, opened 2020-11-28
fix: subtitle CEA708 doesn't show up cause it gets index -1

Yuvalke:

set the index for the native track without language

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -666,6 +666,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         } else if (settings.kind === 'captions' && this._config.enableCEA708Captions) {
           settings.label = settings.label || captionsTextTrackLabels.shift();
           settings.language = settings.language || captionsTextTrackLanguageCodes.shift();
+          settings.emptyLanguage = !!settings.language;
           parsedTracks.push(new PKTextTrack(settings));
         }
       }
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -664,9 +664,9 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         if (settings.kind === 'subtitles') {
           parsedTracks.push(new PKTextTrack(settings));
         } else if (settings.kind === 'captions' && this._config.enableCEA708Captions) {
+          settings.emptyLanguage = !settings.language;
           settings.label = settings.label || captionsTextTrackLabels.shift();
           settings.language = settings.language || captionsTextTrackLanguageCodes.shift();
-          settings.emptyLanguage = !!settings.language;
           parsedTracks.push(new PKTextTrack(settings));
         }
       }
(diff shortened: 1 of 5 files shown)


src/track/text-track.js
```diff
@@ -40,6 +46,15 @@ const TextTrack: TextTrack = class TextTrack extends Track {
     return this._external;
   }
 
+  /**
+   * Getter for the CEA of the text track.
+   * @public
+   * @returns {boolean} - The kind of the text track.
```
RoyBregman, line 52: It is copied pasted wrong :)
```suggestion
   * @returns {boolean} - returns true if the text track has no language specified
```

src/player.js
```diff
@@ -2074,7 +2074,11 @@ export default class Player extends FakeEventTarget {
         const videoElement = this.getVideoElement();
         return videoElement ? Array.from(videoElement.textTracks).findIndex(track => (track ? track.language === textTrack.language : false)) : -1;
       };
-      this._getTextTracks().forEach(track => (track.index = getNativeLanguageTrackIndex(track)));
+      this._getTextTracks().forEach(track => {
+        if (!track.emptyLanguage) {
```
RoyBregman, line 2078: Why do we need this new property called emptyLanguage and the overhead of handing it (all the changes in the TextTrack class). Can't we just check here "if (track.language) {" ?

RoyBregman: Oh, you want it targeted only for the NativeAdapter?

RoyBregman: But I guess this code is correct always... Even if it doesn't happen in other adapters. If there is no language then don't try to look it up.

Yuvalke: no cause it's not empty

Yuvalke: we set value on native adapter for CEA708

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -666,6 +666,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         } else if (settings.kind === 'captions' && this._config.enableCEA708Captions) {
           settings.label = settings.label || captionsTextTrackLabels.shift();
           settings.language = settings.language || captionsTextTrackLanguageCodes.shift();
+          settings.emptyLanguage = !!settings.language;
```
RoyBregman, line 669: Maybe instead we can just update the textTracks on the videoElement, like this:
`            if (!!settings.language) {
              settings.language = settings.language;
            } else {
              textTracks[i].language = settings.language = captionsTextTrackLanguageCodes.shift();
            }`
Then we don't need the new prop and it works perfect.

Yuvalke: you can't update the textTrack on video element it's read only

RoyBregman: But it worked :)

RoyBregman: We manipulate the mode, isn't same to manipulate this also?

Yuvalke: https://developer.mozilla.org/en-US/docs/Web/API/TextTrack

RoyBregman: Yes... You are right. Bummer. Then lets stick with your original solution. Seems to be the best (and only) option

RoyBregman (approved):

RoyBregman:
fix conflicts

PR #514, opened 2020-11-29
fix: CEA708 subtitle doesn't get correct index on native 

Yuvalke:

Issue: addTrack on safari add the track to the first position on TextTrackList.
Solution: register after native tracks loaded to addtexttrack to make sure if new tracks adding on index 0 it'll change the indexes for all tracks otherwise it won't change it.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1747,8 +1747,8 @@ export default class Player extends FakeEventTarget {
     const trackIndex = videoElement
       ? Array.from(videoElement.textTracks).findIndex(track => track && track.language === event.payload.track.language)
       : -1;
-    const textTracks = this._getTextTracks();
     if (trackIndex === 0) {
+      const textTracks = this._getTextTracks();
       // new native track added to start or end of text track list so if it added to start we should fix our indexes
       // by increasing the index by 1
       textTracks.forEach(track => {
--- a/src/player.js
+++ b/src/player.js
@@ -2134,7 +2134,9 @@ export default class Player extends FakeEventTarget {
    */
   _updateTracks(tracks: Array<Track>): void {
     Player._logger.debug('Tracks changed', tracks);
-    this._eventManager.listen(this._engine, CustomEventType.TEXT_TRACK_ADDED, (event: FakeEvent) => this._onTextTrackAdded(event));
+    if (this.config.playback.useNativeTextTrack) {
+      this._eventManager.listen(this._engine, CustomEventType.TEXT_TRACK_ADDED, (event: FakeEvent) => this._onTextTrackAdded(event));
+    }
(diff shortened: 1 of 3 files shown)


src/player.js
```diff
@@ -1736,6 +1736,27 @@ export default class Player extends FakeEventTarget {
...
 
+  /**
+   * The text track changed event object
+   * @param {FakeEvent} event - payload with text track
+   * @returns {void}
+   * @private
+   */
+  _onTextTrackAdded(event: FakeEvent): void {
+    const videoElement = this.getVideoElement();
+    const trackIndex = videoElement
```
dan-ziv, line 1747: 1. This event can be dispatched only if video element is exists, so I don't see a reason for the short-if statement.
2. Even if for some reason the video element isn't exists, `trackIndex` will get `false` value, so it can be boolean or a number? and why to call `getTextTracks` in that scenario? it will be redundant no?
So as I see it, or wrap the whole method logic with `if (videoElement) {...}` or remove the protection.

dan-ziv (changes requested):

src/player.js
Yuvalke: trackIndex fixed to get -1 if no video element.
getTextTracks enter to if condition.
I don't see it as a problem to add this protection in inline if cause it relevant only for this check and next if will ignore it

Yuvalke: trackIndex fixed to get -1 if no video element.
getTextTracks enter to if condition.
I don't see it as a problem to add this protection in inline if cause it relevant only for this check and next if will ignore it.

dan-ziv (approved):

src/player.js
```diff
@@ -1736,6 +1736,27 @@ export default class Player extends FakeEventTarget {
...
   }
 
+  /**
+   * The text track changed event object
+   * @param {FakeEvent} event - payload with text track
+   * @returns {void}
+   * @private
+   */
+  _onTextTrackAdded(event: FakeEvent): void {
+    const videoElement = this.getVideoElement();
```
OrenMe, line 1746: we want to avoid accessing the native video element as much as possible, for example, this will fail for flash and youtube or any other external non HTML5 engine.

Yuvalke: @OrenMe  why? get video element will be null and then nothing happens on the next check, otherwise, text tracks will be empty and if both are not empty for other engines it's ok that the logic will work

OrenMe: Ok, but this logic was not needed till now for non - native text track, so what has changed?
We should strive to get away from this and not add more use cases.

Yuvalke: Flash and youtube won't fire this event anyway

Yuvalke: it's for native text track only, listener added only for `useNativeTextTrack=true`

PR #516, opened 2020-12-07
fix(FEC-10760): Youbora - bufferUnderrun event fired when buffering started and no bufferDuration in Ping events

dan-ziv:

When buffering on live content, the order of events that triggered from the video element is: 
1. `playing`
2. `waiting`
3. `timeupdate`
(this is true for all browsers except IE)
Because we're triggering `playing` on `timeupdate` event we're causing Youbora to report end of buffer when the player is actually still buffering.

The solution is to exclude this logic only to IE.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/state/state-manager.js
+++ b/src/state/state-manager.js
@@ -6,6 +6,7 @@ import {StateType} from './state-type';
 import {CustomEventType, Html5EventType} from '../event/event-type';
 import FakeEvent from '../event/fake-event';
 import getLogger from '../utils/logger';
+import Env from '../utils/env';
 
 /**
  * This class responsible to manage all the state machine of the player.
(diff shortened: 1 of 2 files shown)


src/state/state-manager.js
```diff
@@ -68,87 +69,49 @@ export default class StateManager {
...
-      [Html5EventType.PAUSE]: () => {
-        this._updateState(StateType.PAUSED);
-        this._dispatchEvent();
-      },
+      [Html5EventType.PLAYING]: () => this._updateState(StateType.PLAYING),
+      [Html5EventType.PAUSE]: () => this._updateState(StateType.PAUSED),
       [Html5EventType.TIME_UPDATE]: () => {
-        if (this._player.currentTime !== this._lastWaitingTime && this._prevState && this._prevState.type === StateType.PLAYING) {
+        if (
+          Env.browser.name === 'IE' &&
```
RoyBregman, line 108: So actually this is the only change in logic, right? the rest seems to be refactoring

dan-ziv: Yes...
Just changing the event dispatching to be from `updateState` because you always want to do it and there's no need to split it outside and so many places.

OrenMe (approved):

PR #518, opened 2020-12-08
feat(FEC-10766): create text config section and option for styling

Yuvalke:

Issue: text track doesn't have a config for styling, there is only an API for styling.
Solution: create a text container for the text track and add the options for styling the text track.
Add an option to disable external setting by set attribute explicit as null or ''.
fix test: should dispatch new state event - updateState dispatch the event so two events dispatched cause FEC-10760

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -469,6 +469,7 @@ var config = {
 > {
 >  useNativeTextTrack: boolean,
 >  enableCEA708Captions: boolean,
+>  forceCenter: boolean,
 >  textTrackDisplaySetting: Object,
 >  textStyle: TextStyle,
 >  captionsTextTrack1Label: string,
@@ -484,6 +485,7 @@ var config = {
 > {
 >  useNativeTextTrack: false,
 >  enableCEA708Captions: false,
+>  forceCenter: false,
 >  captionsTextTrack1Label: "English",
 >  captionsTextTrack1LanguageCode: "en",
 >  captionsTextTrack2Label: "Spanish",
@@ -512,6 +514,16 @@ var config = {
 > > ##### Default: `false`
 > >
 > > ##### Description: Whether or not to enable CEA-708 captions.
+>
+> ##
+>
+> > ### config.text.forceCenter
+> >
+> > ##### Type: `Object`
+> >
+> > ##### Default: `false`
+> >
+> > ##### Description: set the forceCenter to true will override the position, align and size in textTrackDisplaySetting
 >
 > ##
 >
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -595,7 +595,8 @@ var config = {
 >  streamPriority: Array<PKStreamPriorityObject>,
 >  preferNative: PKPreferNativeConfigObject,
 >  inBrowserFullscreen: boolean,
->  playAdsWithMSE: boolean
+>  playAdsWithMSE: boolean,
+>  screenLockOrientionMode: string
 > }
 > ```
 >
@@ -614,6 +615,7 @@ var config = {
 >  muted: false,
 >  pictureInPicture: true,
 >  playAdsWithMSE: false,
+>  screenLockOrientionMode: ScreenOrientationType.NONE,
 >  options: {
 >    html5: {
 >      hls: {},
@@ -924,6 +926,20 @@ var config = {
 > > ```
 > >
 > > > ##### Description: Gives the ability to share same video tag to play ads and source with media source
+>
+> ##
+>
+> > ### config.playback.screenLockOrientionMode
+> >
+> > ##### Type: `string` - value list option in ScreenOrientationType
+> >
+> > ##### Default: `none` - ScreenOrientationType.NONE
+> >
+> > ```js
+> > screenLockOrientionMode: string;
+> > ```
+> >
+> > > ##### Description: Gives the ability to lock the screen orientation in fullscreen
 >
 > ##
 >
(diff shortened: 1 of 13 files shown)


src/player.js
```diff
@@ -1248,6 +1249,21 @@ export default class Player extends FakeEventTarget {
...
 
+  _applyTextTrackConfig(): void {
+    if (Utils.Object.getPropertyPath(this._config, 'text.forceCenterCaptions')) {
+      this.setTextDisplaySettings({position: 'auto', align: 'center', size: '100'});
+    } else if (Utils.Object.hasPropertyPath(this._config, 'text.textTrackDisplaySetting')) {
+      this.setTextDisplaySettings(this._config.text.textTrackDisplaySetting);
+    }
+    try {
+      if (Utils.Object.hasPropertyPath(this._config, 'text.textStyle')) {
+        this.textStyle = this._config.text.textStyle;
```
OrenMe, line 1260: need to use the `textStyle` setter as it handles also native text track and in general we should reuse code.
It means you need to create a `TextStyle` instance.
As this is coming from a JSON config we may add to `TextStyle` a static `fromJson` that gets json and returns a populated instance.

flow-typed/types/text-config.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import {TextStyle} from '../../src/track/text-style';
+
+declare type PKTextConfigObject = {
+  enableCEA708Captions: boolean,
+  useNativeTextTrack: boolean,
+  textTrackDisplaySetting: Object,
+  textStyle: TextStyle,
+  forceCenterCaptions: boolean,
```
OrenMe, line 9: I think this will apply to subtitles as well so better to make this more generic and call it `forceCenter`?

docs/configuration.md
```diff
@@ -20,8 +20,7 @@ var player = playkit.core.loadPlayer(config);
   network: PKNetworkConfigObject,
   customLabels: PKCustomLabelsConfigObject,
   abr: PKAbrConfigObject,
-  drm: PKDrmConfigObject,
-  dimensions: PKDimensionsConfig
```
OrenMe, line 24: dod you remove it this on purpose? not used?
With the latest addition by @dan-ziv in #506 I think we can use this as config as well now

OrenMe (changes requested):

docs/configuration.md
Yuvalke: totally by mistake

docs/configuration.md
```diff
@@ -1403,53 +1428,4 @@ var config = {
...
-> >
-> > ##### Default: `''`
-> >
-> > ##### Description: Defines the aspect ratio of the player.
-> >
-> > The aspect ratio should be written in the form of `'width:height'`, for example: `'4:3'` (classic TV ratio).
-> > If one of the `height` or `width` parameters is additionally provided in the configuration, the value of the other parameter not provided will be calculated accordingly to match the aspect ratio. If both were provided, the `height` value would be overridden.
-
-##
-
```
OrenMe, line 1454: also here.

src/player.js
Yuvalke: I used the textStyle setter, as part from https://github.com/kaltura/playkit-js/commit/f06aa5e929f8ab947e10ee3a466a3cc90cd0f62c we can pass instance by config

flow-typed/types/text-config.js
Yuvalke: sounds good

src/player.js
OrenMe: You are right, you use it.
Issue is that when coming from uiconf it won't be instance of TextStyle so we need to normalize it either in Kaltura player when getting this from uiconf or here.

flow-typed/types/text-config.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import {TextStyle} from '../../src/track/text-style';
+
+declare type PKTextConfigObject = {
+  enableCEA708Captions: boolean,
+  useNativeTextTrack: boolean,
+  textTrackDisplaySetting: Object,
+  textStyle: TextStyle | Object,
```
OrenMe, line 8: why Object? we should be strict here. If you want partial then use [$Shape](https://flow.org/en/docs/types/utilities/#toc-shape)

src/player.js
```diff
@@ -1248,6 +1249,21 @@ export default class Player extends FakeEventTarget {
     return false;
   }
 
+  _applyTextTrackConfig(): void {
+    if (Utils.Object.getPropertyPath(this._config, 'text.forceCenter')) {
+      this.setTextDisplaySettings({position: 'auto', align: 'center', size: '100'});
+    } else if (Utils.Object.hasPropertyPath(this._config, 'text.textTrackDisplaySetting')) {
+      this.setTextDisplaySettings(this._config.text.textTrackDisplaySetting);
+    }
```
OrenMe, line 1257: I think this will cause object cache issues.
What happens if someone sets `forceCenter` and then sets `textTrackDisplaySetting`? he will need to remeber to set force off.
They can be set seperatly - if force apply default and if specific apply them, at this order.
Otherwise it requires additional call to configure to reset state or at least adding it in config.

OrenMe (changes requested):

src/player.js
Yuvalke: @OrenMe I think it's a bad configuration, if someone set forceCenter it'll set the `TextDisplaySettings` otherwise his textTrackDisplaySetting config will be applied.
I don't think it's ok to set both, WDYT @OrenMe ?

OrenMe: It is always a problem when you have two settings options that impact same internal config, and in these cases it is best to have it as one config.
Obviously this is not possible here - so we remain in finding the least worse.
If you want something to take precedence then you can apply both but call the force after so it applies.
Use cases that are currently a bit problematic are if app wants to change config it needs to set both configs.
Also, if you want to pass both force and other cue settings which force doesn't config you can't do it in one pass also.

flow-typed/types/text-config.js
Yuvalke: It's not partial, we wanted an option to pass config and create TextStyle instance on the fly, the config is the properties of the object.

src/player.js
Yuvalke: why would you want to pass both? it's not a weird decision? if you have something special then config from TextDisplaySettings otherwise you can use the default for setting cue on center.
if it'll override from the force after you set specific TextDisplaySettings you'll get a different result then you expect on TextDisplaySettings

OrenMe: I do not know the use case as it varies according to specific app requirement.
As it stands now the api is limiting and requires multiple actions to achieve various actions which can be achieved in one call.
The forceCenter is a subset of display settings and can always be applied after it so it takes precedence.
Also if you require setting force and then other attributes that force doesn't affect it won't be possible through studio.

flow-typed/types/text-config.js
OrenMe: Ok, but it should be defined as TextStyle or object with list of keys of TextStyle instead of an open object.

src/player.js
```diff
@@ -1248,6 +1249,26 @@ export default class Player extends FakeEventTarget {
     return false;
   }
 
+  _applyTextTrackConfig(): void {
+    if (Utils.Object.getPropertyPath(this._config, 'text.forceCenter')) {
+      this.setTextDisplaySettings({position: 'auto', align: 'center', size: '100'});
+    }
```
OrenMe, line 1255: let's apply this after `textTrackDisplaySetting` as force will take precedence over any other `position`, `align` and `size`

OrenMe (approved):

PR #520, opened 2020-12-16
fix(FEC-10564): keep the poster for display for an audio-only entry

Yuvalke:

Issue: audio stream has a black screen when it plays.
Solution: check if it's audio by sources.type and keep the poster.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1092,6 +1092,16 @@ export default class Player extends FakeEventTarget {
   }
 
   /**
+   * Checking if the current playback is audio only.
+   * @function isAudio
+   * @returns {boolean} - Whether playback is audio.
+   * @private
+   */
+  isAudio(): boolean {
+    return this._config.sources.type === MediaType.AUDIO;
+  }
+
+  /**
    * Get whether the video is seeked to live edge in dvr
    * @returns {boolean} - Whether the video is seeked to live edge in dvr
    * @public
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -1842,6 +1852,7 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _handleAutoPlay(): void {
+    this._posterManager.show();
```
dan-ziv, line 1855: It won't cause flickering in case of video autoplay?

src/player.js
```diff
@@ -1091,6 +1091,16 @@ export default class Player extends FakeEventTarget {
     );
   }
 
+  /**
+   * Checking if the current playback is audio only.
+   * @function isAudio
+   * @returns {boolean} - Whether playback is audio.
+   * @private
+   */
+  isAudio(): boolean {
```
dan-ziv, line 1100: Add this API in kaltura player?

Yuvalke: It should be added, yes

src/player.js
Yuvalke: I didn't see a flicker

OrenMe: In any case, this is logically not safe - if autoplay is set we shoudn't show a poster.
Need to move to where we know we decided to not show it.

dan-ziv: @Yuvalke try with throttling the browser

OrenMe (approved):

PR #521, opened 2020-12-23
feat(FEC-10709, FEC-10712): player visibility - Auto-pause when player is out of view, Autoplay only when player is in view

RoyBregman:

In continuity of the floating mechanism (Floating Player - Design & Tasks breakdown ) some additional functionality is now 
required to enhance behaviour when player visibility state changes (Due to scrolling / tabbing in/out)

- added "inview" to autoplay options
- added new autopause config to playback
- changed MultiMap to support all types as keys

solves FEC-10709, FEC-10712

### CheckLists

- [ ] Docs have been updated


src/event/event-type.js
```diff
@@ -247,7 +247,19 @@ const CustomEventType: PKEventTypes = {
...
+   */
+  TAB_VISIBILITY_CHANGE: 'tabvisibilitychange',
+  /**
+   * Fired when the player gets in /out of scroll view
+   */
+  SCROLL_VISIBILITY_CHANGE: 'scrollvisibilitychange',
+  /**
+   * Fired when the player changes its visibility state - either by tabbing in / out or scrolling in /out of view
+   */
+  VISIBILITY_CHANGE: 'visibilitychange'
```
OrenMe, line 262: I still don't see why we need to add seperate events, all say the same at the end of the day.
Either check with Kava if tab is same as scroll - i believe it does, or add the change to include a reason with tab/scroll - which I think is not really needed.

src/auto-play-type.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+const AutoPlayType: PKAutoPlayTypes = {
+  true: true,
+  IN_VIEW: 'inview'
+};
```
OrenMe, line 5: false?

src/player.js
```diff
@@ -399,6 +407,47 @@ export default class Player extends FakeEventTarget {
...
+   * @private
+   */
+  _isVisible: boolean;
+
+  /**
+   * Whether the player is in floating mode
+   * @type {boolean}
+   * @private
+   */
+  _isFloating: boolean = false;
```
OrenMe, line 443: if core player is aware of floating state which is from a plugin we are breaking isolation of features

src/player.js
```diff
@@ -170,12 +171,19 @@ export default class Player extends FakeEventTarget {
...
+   */
+  _mediaEventManager: EventManager;
   /**
    * The event manager of the player.
    * @type {EventManager}
    * @private
    */
-  _eventManager: EventManager;
+  _playerEventManager: EventManager;
+
```
OrenMe, line 186: why two seperate event managers?

OrenMe (changes requested):
Copnsider seperating the entire visibility managment to seperate manager in its own module.

src/event/event-type.js
RoyBregman: For example the floating feature is only interested of scrolling in / out and not tabbing in/out.
I guess other features might be interested of just tabbing.
I don't see a reason why not giving all options open

src/auto-play-type.js
RoyBregman: I added

src/player.js
RoyBregman: Okay... But this is part of the product request...
How else can I implement this?
Why is this breaking isolation? I am not relying of the plugin to be part. Either it is and then I consider it or not.

src/event/event-type.js
OrenMe: So let's use reason

src/player.js
RoyBregman: it is like saying not to check PIP because it is implemented by the browser.

src/player.js
RoyBregman: The current event manager we have is cleaned and attached every media switch.
We need an event manager that is kept through all the player instance.

src/player.js
OrenMe: Core doesn't know about floating so now it needs to.
If you want to have some generic way to signal core player to ignore condition then this is a generic api that can then be used for other plugins if they want to tell the feature to not trigger

src/player.js
OrenMe: So see comment about separating to its own manager, we do same in other managers.

src/player.js
OrenMe: And per pip, it is not the same as we implement pip internally while floating is not

src/player.js
```diff
@@ -1897,6 +2059,19 @@ export default class Player extends FakeEventTarget {
...
   }
 
+  /**
+     }
+   * Checks auto play configuration and handles initialization accordingly.
+   * @returns {void}
+   * @private
+   */
+  _initAutoPlay(): void {
+    this._posterManager.show();
```
yairans, line 2069: why you changed the previous logic which shows the poster only for autoplay false?
maybe it could cause a flicker

src/player.js
```diff
@@ -1725,6 +1863,16 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  /**
+   * Update the floating active state.
+   * @param {boolean} isFloating - is player floating.
+   * @returns {void}
+   * @public
+   */
+  changeFloatingState(isFloating: boolean): void {
```
yairans, line 1872: `toggleFloatingState` ? or just `floatingState` setter?

src/player.js
```diff
@@ -1837,33 +1985,47 @@ export default class Player extends FakeEventTarget {
...
-            }
-          } else {
-            onAutoPlayFailed();
+  _initAutoPause(): void {
+    if (this._config.playback.autopause === true) {
+      this._mediaEventManager.listen(this, CustomEventType.VISIBILITY_CHANGE, (e: FakeEvent) => {
+        if (!e.payload.visible) {
+          if (!this._isFloating && !this.isInPictureInPicture() && this._playbackStart && !this.paused) {
+            this.pause();
+            this._autoPaused = true;
```
yairans, line 1998: I would reset it in playing for when the play and pause called by API
(paused on invisible => played by api => paused by api => become visible)

src/player.js
```diff
@@ -399,6 +407,47 @@ export default class Player extends FakeEventTarget {
...
+   * @private
+   */
+  _isTabVisible: boolean;
+
+  /**
+   * Whether the player browser tab is active and in the scroll view
+   * @type {boolean}
+   * @private
+   */
+  _isVisible: boolean;
```
yairans, line 436: is not `_isVisibleInScroll && _isTabVisible`?

src/playkit.js
```diff
@@ -33,6 +33,8 @@ import {AdBreakType} from './ads/ad-break-type';
 import {AdTagType} from './ads/ad-tag-type';
 import {AdEventType} from './ads/ad-event-type';
 import {ScreenOrientationType} from './screen-orientation-type';
+import 'intersection-observer';
```
yairans, line 36: why it's needed?

yairans (changes requested):

src/player.js
Yuvalke: and add also isAudio to this condition, the same as we have  in the master today

src/utils/viewability-manager.js
```diff
@@ -0,0 +1,55 @@
...
+   * @param {Function} listener - the callback to be invoked when visibility is changed (and when starting to observe). The callback is called with a boolean param representing the visibility state
+   * @returns {void}
+   */
+  observe(target: HTMLElement, threshold: number, listener: Function): void {
+    const options = {
+      threshold: threshold
+    };
+    const intersectionObserver = new window.IntersectionObserver((entries: Array<window.IntersectionObserverEntry>) => {
+      const visible = entries[0].intersectionRatio >= threshold;
+      (listener: Function).call(this, visible);
```
Yuvalke, line 25: why don't just call the listener?

src/player.js
```diff
@@ -1528,6 +1574,83 @@ export default class Player extends FakeEventTarget {
...
+  get isTabVisible(): boolean {
+    return this._isTabVisible;
+  }
+
+  /**
+   * Gets the player visibility in scroll view state
+   * @returns {boolean} - whether the player is visible in the scroll view
+   * @public
+   */
+  get isVisibleInScroll(): boolean {
```
Yuvalke, line 1642: why do you need the getter and _isVisibleInScroll if you don't have the event SCROLL_VISIBILITY_CHANGE anymore?

Yuvalke: why do we need those getters? user can use events for this state changes

Yuvalke (changes requested):

src/player-config.js
```diff
@@ -58,7 +58,8 @@ const DefaultConfig = {
         engine: 'flash',
         format: 'hls'
       }
-    ]
+    ],
+    visibilityThreshold: 50
```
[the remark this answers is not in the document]
Yuvalke: I'm not sure if playback is the right place for this config

Yuvalke (changes requested):

src/player-config.js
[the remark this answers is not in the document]
RoyBregman: We talked about it at the DR, you are right, but do you have a better suggestion?

src/player.js
RoyBregman: the getter isVisibleInScroll is really not relevant anymore. I removed it.
The other getters are needed to get the initial state

Yuvalke: the initial state won't be always yes?

src/event/event-type.js
RoyBregman: Not relevant anymore. Code moved to Kaltura. Only new event will be dispatched from Kaltura - VISIBILITY_CHANGE

src/player.js
RoyBregman: Not relevant anymore. Code was refactored

src/player.js
RoyBregman: Not relevant anymore. Reverted back to one manager

src/auto-play-type.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+const AutoPlayType: PKAutoPlayTypes = {
+  true: true,
```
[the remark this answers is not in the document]
RoyBregman: Fixed and change TRUE to ALWAYS

src/event/event-manager.js
```diff
@@ -12,7 +12,7 @@ import FakeEvent from './fake-event';
  * @implements {IDestroyable}
  */
 class EventManager {
-  _bindingMap: MultiMap<Binding_> | null;
+  _bindingMap: MultiMap<string, Binding_> | null;
```
Yuvalke, line 15: why did you change it?

Yuvalke (changes requested):

Yuvalke:
Please change the PR comment and run docs:generate cause you added a new API

src/player.js
```diff
@@ -171,11 +172,12 @@ export default class Player extends FakeEventTarget {
   }
 
   /**
-   * The event manager of the player.
+   * The event manager of the current media.
```
OrenMe, line 175:
```suggestion
   * The event manager of the player.
```

src/player.js
```diff
@@ -461,7 +463,7 @@ export default class Player extends FakeEventTarget {
         this._posterManager.setSrc(this._config.sources.poster);
         this._handleDimensions();
         this._handlePreload();
-        this._handleAutoPlay();
+        this._initAutoPlay();
```
OrenMe, line 466: if we don't change anything let's keep current name

src/auto-play-type.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+const AutoPlayType: PKAutoPlayTypes = {
+  ALWAYS: true,
```
OrenMe, line 3: keep it true, as we don't call false NEVER

src/event/event-type.js
```diff
@@ -251,7 +251,11 @@ const CustomEventType: PKEventTypes = {
   /**
    * Fired when the drm license is responded from the DRM server
    */
-  DRM_LICENSE_LOADED: 'drmlicenseloaded'
+  DRM_LICENSE_LOADED: 'drmlicenseloaded',
+  /**
+   * Fired when the player changes its visibility state - either by tabbing in / out or scrolling in /out of view
+   */
+  VISIBILITY_CHANGE: 'visibilitychange'
```
OrenMe, line 258: why do we need this here? we can pull it to kaltura-player and merge it in the events getter

src/player-config.js
```diff
@@ -78,7 +78,11 @@ const DefaultConfig = {
   drm: {
     keySystem: ''
   },
-  network: {}
+  network: {},
+  visibility: {
+    tolerance: 10,
+    playerThreshold: 50
+  }
```
OrenMe, line 85: move to kaltura-player

src/player-config.js
```diff
@@ -78,7 +78,11 @@ const DefaultConfig = {
   drm: {
     keySystem: ''
   },
-  network: {}
+  network: {},
+  visibility: {
+    tolerance: 10,
```
OrenMe, line 83: i think this name is not to clear, and it will be more correct to get this as an array cause sometimes people will want non linear stops

src/player.js
```diff
@@ -1902,35 +1908,25 @@ export default class Player extends FakeEventTarget {
...
-        } else {
-          if (capabilities.mutedAutoPlay) {
-            if (this.muted && !this._fallbackToMutedAutoPlay) {
-              onMutedAutoPlay();
-            } else if (allowMutedAutoPlay) {
-              onFallbackToMutedAutoPlay();
-            } else {
-              onAutoPlayFailed();
-            }
+  autoPlay(): void {
```
OrenMe, line 1911: it is not correct to call this autoPlay, as it is more correct to say this is a non user gesture play AKA programmatic play.
WDYT to put this in the play API and pass to it an object {programmatic: boolean} and if it is yes then call `_handleAutoPlay` and return.
BTW, the check for audio should still show the poster in this case.

src/auto-play-type.js
RoyBregman: We could call it OFF but ok, reverted to TRUE :)

src/player.js
OrenMe: @RoyBregman ? is this change necessary?

src/player.js
OrenMe: @RoyBregman ?

src/player.js
RoyBregman: No, it is not, reverted back to original name

src/player-config.js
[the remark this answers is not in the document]
RoyBregman: fixed

src/player.js
RoyBregman: fixed

src/player.js
RoyBregman: not relevant anymore

src/player.js
RoyBregman: not relevant anymore

src/playkit.js
RoyBregman: ie11 support. moved to viewabilityManager

src/utils/viewability-manager.js
RoyBregman: fixed

src/player.js
RoyBregman: Not relevant anymore

src/event/event-type.js
RoyBregman: done

src/player-config.js
RoyBregman: done

src/player-config.js
RoyBregman: Not relevant anymore

src/player.js

src/player.js
```diff
@@ -517,10 +517,16 @@ export default class Player extends FakeEventTarget {
 
   /**
    * Start/resume playback.
+   * @param {boolean} programmatic - indicates the play call was not user gestured and should be handled like auto play.
    * @returns {void}
    * @public
    */
-  play(): void {
+  play(programmatic: boolean = false): void {
```
OrenMe, line 524: I think adding a true/false boolean to API is not clear, this is why I suggested to add it as an object so when one calls the API it is clear what is the action: `player({programmatic: true})`

src/player.js
OrenMe: please see https://github.com/kaltura/playkit-js/pull/521#discussion_r551856918 - this comment is still relevant and not handled.

> BTW, the check for audio should still show the poster in this case.

was this checked? I guess it should be OK as on initial load if it is audio we will get here and show it - please make sure this is covered and make sure it is a tested use case in QA plan

RoyBregman: I'm not sure I understand the comment - audio case logic is remained the same. So is autoplay true and false.
The only change in logic here is that I show poster also for autoplay "inview".
This is done because someone can define threshold as 90% and scroll slowly until he reaches that point. We don't want to see a black player in the mean while.

src/player.js
```diff
@@ -517,12 +517,12 @@ export default class Player extends FakeEventTarget {
 
   /**
    * Start/resume playback.
-   * @param {boolean} programmatic - indicates the play call was not user gestured and should be handled like auto play.
+   * @param {PKPlayOptionsObject} playOptions - additional options to control the play. i.e. programmatic true indicates the play call was not initiated by a user gesture and should be handled like auto play.
```
OrenMe, line 520: see example on hot to describe object properties with jsdoc https://stackoverflow.com/questions/6460604/how-to-describe-object-arguments-in-jsdoc

OrenMe (approved):

flow-typed/types/auto-play-types.js
```diff
@@ -0,0 +1,2 @@
+// @flow
+declare type PKAutoPlayTypes = { [type: string | boolean]: string | boolean };
```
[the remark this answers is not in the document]
RoyBregman: yes, fixed

src/player.js
```diff
@@ -1837,33 +1985,47 @@ export default class Player extends FakeEventTarget {
...
-              onMutedAutoPlay();
-            } else if (allowMutedAutoPlay) {
-              onFallbackToMutedAutoPlay();
-            } else {
-              onAutoPlayFailed();
-            }
-          } else {
-            onAutoPlayFailed();
+  _initAutoPause(): void {
+    if (this._config.playback.autopause === true) {
```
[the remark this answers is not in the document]

src/player.js
RoyBregman: Not relevant anymore

OrenMe (approved):

OrenMe (dismissed):

OrenMe (approved):

Yuvalke (approved):

PR #525, opened 2020-12-30
fix(FEC-10844): advanced caption settings custom caption is always marked even when choose sample

Yuvalke:

Issue: the condition of || doesn't work properly with 0 as the current value and always takes the default option.
Solution: create the default values as a static getter to be able to use it on static methods and on the first initialize of the properties.
merge the setting and default value to make sure all set correctly.
add relevant tests.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/track/text-style.js
+++ b/src/track/text-style.js
@@ -1,6 +1,4 @@
 // @flow
-import * as Utils from '../utils/util';
-
 /**
  * We use this number to calculate the scale of the text. so it will be : 1 + 0.25 * FontSizes.value
  * So, if the user selects 400% the scale would be: 1 + 0.25 * 4 = 2. so the font size should be multiplied by 2.
@@ -140,30 +138,23 @@ class TextStyle {
     return 'rgba(' + color.concat(opacity).join(',') + ')';
   }
 
-  static get defaultValues(): PKTextStyleObject {
-    return {
-      fontEdge: TextStyle.EdgeStyles.NONE,
-      fontSize: '100%',
-      fontScale: 1,
-      fontColor: TextStyle.StandardColors.WHITE,
-      fontOpacity: TextStyle.StandardOpacities.OPAQUE,
-      backgroundColor: TextStyle.StandardColors.BLACK,
-      backgroundOpacity: TextStyle.StandardOpacities.OPAQUE,
-      fontFamily: TextStyle.FontFamily.SANS_SERIF
-    };
-  }
-
   static fromJson(setting: PKTextStyleObject): TextStyle {
-    const currentSetting = Utils.Object.mergeDeep(TextStyle.defaultValues, setting);
+    const getValue = (newValue, defaultValue) => {
+      let value = defaultValue;
+      if (typeof newValue !== 'undefined' && newValue !== null) {
+        value = newValue;
+      }
+      return value;
+    };
     let textStyle = new TextStyle();
-    textStyle.fontEdge = currentSetting.fontEdge;
-    textStyle.fontSize = currentSetting.fontSize;
-    textStyle.fontScale = currentSetting.fontScale;
-    textStyle.fontColor = currentSetting.fontColor;
-    textStyle.fontOpacity = currentSetting.fontOpacity;
-    textStyle.backgroundColor = currentSetting.backgroundColor;
-    textStyle.backgroundOpacity = currentSetting.backgroundOpacity;
-    textStyle.fontFamily = currentSetting.fontFamily;
+    textStyle.fontEdge = getValue(setting.fontEdge, textStyle.fontEdge);
+    textStyle.fontSize = getValue(setting.fontSize, textStyle.fontSize);
+    textStyle.fontScale = getValue(setting.fontScale, textStyle.fontScale);
+    textStyle.fontColor = getValue(setting.fontColor, textStyle.fontColor);
+    textStyle.fontOpacity = getValue(setting.fontOpacity, textStyle.fontOpacity);
+    textStyle.backgroundColor = getValue(setting.backgroundColor, textStyle.backgroundColor);
+    textStyle.backgroundOpacity = getValue(setting.backgroundOpacity, textStyle.backgroundOpacity);
+    textStyle.fontFamily = getValue(setting.fontFamily, textStyle.fontFamily);
     return textStyle;
   }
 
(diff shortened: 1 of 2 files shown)


src/track/text-style.js
```diff
@@ -138,17 +140,31 @@ class TextStyle {
...
+    let textStyle = new TextStyle();
+    textStyle.fontEdge = currentSetting.fontEdge;
+    textStyle.fontSize = currentSetting.fontSize;
+    textStyle.fontScale = currentSetting.fontScale;
+    textStyle.fontColor = currentSetting.fontColor;
+    textStyle.fontOpacity = currentSetting.fontOpacity;
+    textStyle.backgroundColor = currentSetting.backgroundColor;
+    textStyle.backgroundOpacity = currentSetting.backgroundOpacity;
+    textStyle.fontFamily = currentSetting.fontFamily;
+    return textStyle;
```
OrenMe, line 167: In general I think the better approach here is to use a helper method to check if settings exist and not null or undefined:
```javascript
function get(value, default) {
  let result = default;
  if (value && value !== null && value !== undefined) {
    result = value;
  }
  return result;
}


let textStyle = new TextStyle();
textStyle.fontEdge = get(setting.fontEdge, clonedTextStyle.fontEdge);
textStyle.fontSize = get(setting.fontSize, clonedTextStyle.fontSize);
textStyle.fontScale = get(setting.fontScale, clonedTextStyle.fontScale);
textStyle.fontColor = get(setting.fontColor, clonedTextStyle.fontColor);
textStyle.fontOpacity = get(setting.fontOpacity, clonedTextStyle.fontOpacity);
textStyle.backgroundColor = get(setting.backgroundColor, clonedTextStyle.backgroundColor);
textStyle.backgroundOpacity = get(setting.backgroundOpacity, clonedTextStyle.backgroundOpacity);
textStyle.fontFamily = get(setting.fontFamily, clonedTextStyle.fontFamily;
return textStyle;

```

and if we take it further, there's a new JS feature called [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) that does exactly this:

```javascript
textStyle.fontEdge = setting.fontEdge ?? clonedTextStyle.fontEdge;
```

and there's a babel plugin for it: @babel/plugin-proposal-nullish-coalescing-operator

OrenMe (changes requested):

src/track/text-style.js
```diff
@@ -138,17 +140,31 @@ class TextStyle {
     return 'rgba(' + color.concat(opacity).join(',') + ')';
   }
 
+  static get defaultValues(): PKTextStyleObject {
```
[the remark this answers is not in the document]
Yuvalke: why not? if the user wants to know what is the default value?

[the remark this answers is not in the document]
OrenMe: There are too many options to get default value. You can also create a new textStyle and get value from it.
If you have a justified reason for new api feel free to raise it.

src/track/text-style.js
Yuvalke: this one looks better to you?? I thought it will be better to merge cause it will take the value if exists

OrenMe: If all of the ECMA script experts that write the JavaScript spec think this is good and this is a supported feature already in browsers(check chrome dev tools now if you want 😀 ) then it is good enough to me.

Yuvalke: I mean the get helper :)

src/track/text-style.js
```diff
@@ -139,16 +139,23 @@ class TextStyle {
...
-    clonedTextStyle.backgroundColor = setting.backgroundColor || clonedTextStyle.backgroundColor;
-    clonedTextStyle.backgroundOpacity = setting.backgroundOpacity || clonedTextStyle.backgroundOpacity;
-    clonedTextStyle.fontFamily = setting.fontFamily || clonedTextStyle.fontFamily;
-    return clonedTextStyle;
+    const getValue = (newValue: any, defaultValue: any): any => {
+      let value = defaultValue;
+      if (typeof newValue !== 'undefined' && newValue !== null) {
+        value = newValue;
+      }
+      return value;
```
RoyBregman, line 147: can be replace with a one liner:
 `return (typeof newValue !== 'undefined' && newValue !== null) ? newValue : defaultValue`
 
 But its a matter of taste...

src/track/text-style.js
```diff
@@ -139,16 +139,23 @@ class TextStyle {
...
+    };
+    let textStyle = new TextStyle();
+    textStyle.fontEdge = getValue(setting.fontEdge, textStyle.fontEdge);
+    textStyle.fontSize = getValue(setting.fontSize, textStyle.fontSize);
+    textStyle.fontScale = getValue(setting.fontScale, textStyle.fontScale);
+    textStyle.fontColor = getValue(setting.fontColor, textStyle.fontColor);
+    textStyle.fontOpacity = getValue(setting.fontOpacity, textStyle.fontOpacity);
+    textStyle.backgroundColor = getValue(setting.backgroundColor, textStyle.backgroundColor);
+    textStyle.backgroundOpacity = getValue(setting.backgroundOpacity, textStyle.backgroundOpacity);
+    textStyle.fontFamily = getValue(setting.fontFamily, textStyle.fontFamily);
```
RoyBregman, line 157: Can this maybe replaced with an iteration?

test/src/player.spec.js
```diff
@@ -1216,14 +1216,46 @@ describe('Player', function () {
...
-        textStyle.backgroundColor = TextStyle.StandardColors.RED;
-        textStyle.fontColor = TextStyle.StandardColors.CYAN;
         textStyle.fontEdge = TextStyle.EdgeStyles.RAISED;
+        textStyle.fontSize = '75%';
+        textStyle.fontScale = '3';
+        textStyle.fontColor = TextStyle.StandardColors.BLACK;
+        textStyle.fontOpacity = TextStyle.StandardOpacities.SEMI_HIGH;
+        textStyle.backgroundOpacity = TextStyle.StandardOpacities.SEMI_LOW;
+        textStyle.fontFamily = TextStyle.FontFamily.ARIAL;
+        textStyle.backgroundColor = TextStyle.StandardColors.RED;
```
RoyBregman, line 1226: one of these tests is with 0 value?
the black color?

Yuvalke: Transperncy

Yuvalke: I used all the options to make sure I didn't miss something

src/track/text-style.js
Yuvalke: if we'll get a new non-relevant key/private key it'll add to this set.

src/track/text-style.js
RoyBregman: So unless you wanna change this to my suggestion then it is approved

RoyBregman (approved):

PR #530, opened 2021-01-13
fix(REACH2-919): correct order of captions that has the same time

semarche-kaltura:

For captions with the same time player renders subtitles in wrong order. (https://kaltura.atlassian.net/browse/REACH2-919)
Current fix change the order of iteration and subtitles displays correctly.

Screen Shot 2021-01-12 at 11 56 04 AM

Screen Shot 2021-01-12 at 11 55 41 AM

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/track/text-track-display.js
+++ b/src/track/text-track-display.js
@@ -1036,7 +1036,7 @@ function processCues(window, cues, overlay, style) {
   (function () {
     let styleBox, cue;
 
-    for (let i = 0; i < cues.length; i++) {
+    for (let i = cues.length - 1; i >= 0; i--) {
       cue = cues[i];
 
       // Compute the intial position and styles of the cue div.


RoyBregman:
check with Oren, but i think it is better to open a FEC ticket and link them and solve this as part of FEC project

yairans (approved):

OrenMe (changes requested):
I want to make sure we are actually doing the correct change.
Need to check in spec if it is legal to have such format of separate VTT cues at the exact same times.
The proper solution is to combine them and manage the ordering at the same place.
AFAICS this is the real order of the cues as inserted by the engine.

A good place to compare is to try this in Safari and see how it handles it.

OrenMe:
this is a marginal issue, if there are captions at same time the WebVTT spec doesn't mandate what order they should be stacked and in general for such a use case you would expect the captions to be merged.
Another approach is by defining the line or region for the captions - although we do not currently support VTT regions.

PR #533, opened 2021-01-18
fix(FEC-10890): the captions aren't displayed for audio entry and UI doesn't clickable.

Yuvalke:

Issue: subtitles container append before poster and located under it on the DOM tree.
Solution: poster, captions, and black screen container blocking user action from UI add pointer-events: none to avoid it.
append captions container after playkit containers.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1595,11 +1595,6 @@ export default class Player extends FakeEventTarget {
    * @returns {void}
    */
   _appendDomElements(): void {
-    // Append playkit-subtitles
-    this._textDisplayEl = Utils.Dom.createElement('div');
-    Utils.Dom.setAttribute(this._textDisplayEl, 'aria-live', 'polite');
-    Utils.Dom.addClassName(this._textDisplayEl, SUBTITLES_CLASS_NAME);
-    Utils.Dom.appendChild(this._el, this._textDisplayEl);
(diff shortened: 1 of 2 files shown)


src/player.js
```diff
@@ -1608,6 +1603,11 @@ export default class Player extends FakeEventTarget {
     const el = this._posterManager.getElement();
     Utils.Dom.addClassName(el, POSTER_CLASS_NAME);
     Utils.Dom.appendChild(this._el, el);
+    // Append playkit-subtitles
+    this._textDisplayEl = Utils.Dom.createElement('div');
+    Utils.Dom.setAttribute(this._textDisplayEl, 'aria-live', 'polite');
+    Utils.Dom.addClassName(this._textDisplayEl, SUBTITLES_CLASS_NAME);
+    Utils.Dom.appendChild(this._el, this._textDisplayEl);
```
yairans, line 1610: @Yuvalke what if there is captions at 0 time and preload true. the captions will displayed over the pre-playback screen no?

Yuvalke: Nothing change for pre-playback it's internal change on playkit on for the order it appends on.
Anyway pre-playback has z-index:1

yairans (approved):

PR #536, opened 2021-01-21
fix(FEC-9370): safari video starts from beginning, not from startTime

Yuvalke:

Issue: there are videos like ima dai sample that the video tag fires duration change twice and disable the seek action because of that.
Solution: for every duration change(should be 1 for media) we'll make the seek, for this media, it'll make the seek after the second duration change and then we'll be in the correct place.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -444,8 +444,15 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
         this._handleLiveDurationChange();
       }
     };
-    if (startTime && startTime > -1) {
-      this._videoElement.currentTime = startTime;
+    const setStartTime = seekTo => {
+      if (seekTo && seekTo > -1) {
+        this._videoElement.currentTime = seekTo;
+      }
+    };
+    setStartTime(startTime);
+    if (!this.isLive()) {
+      //for stream that have two duration change - like ima dai sample
+      this._eventManager.listen(this._videoElement, Html5EventType.DURATION_CHANGE, () => setStartTime(startTime));
     }
     if (this._videoElement.textTracks.length > 0) {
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -444,8 +444,15 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
-      this._videoElement.currentTime = startTime;
+    const setStartTime = seekTo => {
+      if (seekTo && seekTo > -1) {
+        this._videoElement.currentTime = seekTo;
+      }
+    };
+    setStartTime(startTime);
+    if (!this.isLive()) {
+      //for stream that have two duration change - like ima dai sample
+      this._eventManager.listen(this._videoElement, Html5EventType.DURATION_CHANGE, () => setStartTime(startTime));
```
OrenMe, line 455: do we understand why this is happening? this looks like an issue in the specific media tested rather than an issue with player.
What is the duration of the video after the first durationchange event? is it different from the second one? please note that DAI playlists are stitched playlists so maybe this is the issue

Yuvalke: OK, Yes the duration is different in milliseconds and it comes from the difference between the video and audio segments. @OrenMe does it OK that there is a difference? or an issue with the stream? I found out different IMA DAI media that doesn't have this difference and it works well. Not sure if we want it in this case.

OrenMe: So yes, this is issue in stream.
I think it is ok to reject this and raise a bug with IMA with this issue, rather than adding fixes for invalid cases.
These kind of issues can also break the video ended events in some cases.

Yuvalke: Agree

PR #543, opened 2021-02-16
feat(FEC-10768): expose in-stream DASH thumbnails

dan-ziv:

* Add definition of `ImageTrack`.
* Add definition of `ThumbnailInfo`.
* Add new API `getThumbnail(time)` to player, engine and adapters interface.
* Move all types to dedicated `enums` folder for better folder structure.

### CheckLists

- [ ] Docs have been updated


src/thumbnail/thumbnail-info.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+class ThumbnailInfo {
+  _url: string;
+  _width: number;
+  _height: number;
+  _x: number;
+  _y: number;
+
+  set url(value: string) {
```
yairans, line 9: better to set all in the constructor and expose gets only

yairans (changes requested):

src/thumbnail/thumbnail-info.js
dan-ziv: Prefer to keep it like that

yairans: can you explain? as I see in dash and kaltura-player, they can pass these params to the ctor 
why not keep this api read only?

yairans (approved):

PR #544, opened 2021-02-23
fix(FEC-11037): multiple decorator exist after destroy plugin with decorator

Yuvalke:

Issue: when the plugin has the decorator and the player destroyed the decorator reregister.
Solution: filter out the previous decorator provider and keep the new one only.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/flow-typed/interfaces/engine-decorator-provider.js
+++ b/flow-typed/interfaces/engine-decorator-provider.js
@@ -1,6 +1,0 @@
-// @flow
-import type {IEngine} from './engine';
-
-declare interface IEngineDecoratorProvider {
-  getEngineDecorator(engine: IEngine, dispatchEventHandler: Function): IEngineDecorator;
-}
--- a/src/engines/engine-decorator-manager.js
+++ b/src/engines/engine-decorator-manager.js
@@ -8,9 +8,9 @@ import {MultiMap} from '../utils';
 class EngineDecoratorManager {
   _decoratorProviders: MultiMap<string, IEngineDecoratorProvider> = new MultiMap();
 
-  register(name: string, engineDecoratorProvider: IEngineDecoratorProvider): void {
-    if (!this._decoratorProviders.has(name)) {
-      this._decoratorProviders.push(name, engineDecoratorProvider);
+  register(engineDecoratorProvider: IEngineDecoratorProvider): void {
+    if (!this._decoratorProviders.has(engineDecoratorProvider.getName())) {
+      this._decoratorProviders.push(engineDecoratorProvider.getName(), engineDecoratorProvider);
     }
   }
 
--- a/src/engines/engine-decorator-manager.js
+++ b/src/engines/engine-decorator-manager.js
@@ -1,22 +1,25 @@
 // @flow
+import {MultiMap} from '../utils';
 
 /**
  * Engine decorator manager for plugins.
  * @class EngineDecoratorManager
  */
 class EngineDecoratorManager {
-  _decoratorGenerators: Array<EngineDecoratorGenerator> = [];
+  _decoratorProviders: MultiMap<string, IEngineDecoratorProvider> = new MultiMap();
 
-  addDecorator(decoratorGenerator: EngineDecoratorGenerator): void {
-    this._decoratorGenerators.push(decoratorGenerator);
+  register(name: string, engineDecoratorProvider: IEngineDecoratorProvider): void {
+    if (!this._decoratorProviders.has(name)) {
+      this._decoratorProviders.push(name, engineDecoratorProvider);
+    }
   }
 
   createDecorators(engine: IEngine, dispatchEvent: Function): Array<IEngineDecorator> {
-    return this._decoratorGenerators.map(decoratorGenerator => decoratorGenerator(engine, dispatchEvent));
+    return this._decoratorProviders.getAll().map(engineDecoratorProvider => engineDecoratorProvider.getEngineDecorator(engine, dispatchEvent));
   }
 
   destroy(): void {
-    this._decoratorGenerators = [];
+    this._decoratorProviders.clear();
   }
 }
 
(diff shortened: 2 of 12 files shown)


src/engines/engine-decorator.js
```diff
@@ -17,9 +17,10 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
 
   static register(decoratorProvider: IEngineDecoratorProvider): void {
     if (decoratorProvider) {
-      if (!EngineDecorator._decoratorProviders.includes(decoratorProvider)) {
-        EngineDecorator._decoratorProviders.push(decoratorProvider);
-      }
+      EngineDecorator._decoratorProviders = EngineDecorator._decoratorProviders.filter(
+        currentDecoratorProvider => currentDecoratorProvider.constructor.name !== decoratorProvider.constructor.name
```
dan-ziv, line 21: Why name and not id?

Yuvalke: we're not maintaining id field

OrenMe:
Hi @yuval, not sure I fully understand the issue description

Yuvalke:
@OrenMe every setup register decorator provider, after destroy we setup again so if plugin has decorator it will add new decorator provider for each destroy and step and create decorator for each provider, this change will keep the newest plugin instance - decorator provider and remove the older one

OrenMe:
So sounds like we need to fix the registry mechanism to be by name or some hash instead of function and then we can actually check that a decorator was registered.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Yuvalke:
@OrenMe static hash? name? I think this check is much better

OrenMe:
I think the entire solution we have today is not ok - it was made when plugins were manager by the playkit lib and not relevant for kaltura player managed plugins.
We register the entire plugin instance in order to just get the decorator from the `getEngineDecorator` method.
Beside the fact we pass the entire plugin which might cause potential memory leaks, if we ever want to get the registered decorators it is impossible with current implementation.

Yuvalke:
@OrenMe agrees, but in our case, we have to fix multiple decorators after destroy.
I didn't understand why it was OK on playkit and doesn't on kaltura?

OrenMe:
The comment about being ok or not was referring to passing plugin instance from Kaltura player to Playkit.
When playkit managed the plugin instance it made sense, now Kaltura player does it so it makes no sense to pass the entire plugin instance to it just in order for it to call the getDecorator method.

src/engines/engine-decorator-manager.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+
+/**
+ * Engine decorator manager for plugins.
+ * @class EngineDecoratorManager
+ */
+class EngineDecoratorManager {
+  _decoratorGenerators: Array<EngineDecoratorGenerator> = [];
+
+  addDecorator(decoratorGenerator: EngineDecoratorGenerator): void {
```
OrenMe, line 10:
```suggestion
  register(name: string, engineDecoratorProvider: EngineDecoratorProvider): void {
```

src/engines/engine-decorator-manager.js
```diff
@@ -0,0 +1,23 @@
...
+    this._decoratorGenerators.push(decoratorGenerator);
```
OrenMe, line 11: add check to see if it is already provided, and with above addition of name it is easier then ever

OrenMe (changes requested):
General note - most changes are around naming conventions or keeping existing names

flow-typed/interfaces/engine-decorator-provider.js
```diff
@@ -1,6 +0,0 @@
-// @flow
-import type {IEngine} from './engine';
-
-declare interface IEngineDecoratorProvider {
-  getEngineDecorator(engine: IEngine, dispatchEventHandler: Function): IEngineDecorator;
```
OrenMe, line 5: this needs to be part of the plugin interface, it is ok to remove it from here but need to make sure we define it for plugins in kaltura-player-js

Yuvalke: Sure the change for kaltura player didn't upload yet

src/engines/engine-decorator-manager.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import {MultiMap} from '../utils';
+
+/**
+ * Engine decorator manager for plugins.
+ * @class EngineDecoratorManager
+ */
+class EngineDecoratorManager {
+  _decoratorProviders: MultiMap<string, IEngineDecoratorProvider> = new MultiMap();
```
OrenMe, line 9: this can be a simple map, multi map is were the value is array but we only have one decorator per plugin

src/player.js
```diff
@@ -390,6 +391,12 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _aspectRatio: ?string;
+  /**
+   * The engine manager.
```
OrenMe, line 395:
```suggestion
   * The engine decorator manager.
```

src/engines/engine-decorator-provider.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+/**
+ * Engine decorator provider.
+ * @class EngineDecoratorProvider
+ * @param {IEngineDecoratorProvider} plugin - The plugin which have implemented decorator.
+ * @implements {IEngineDecorator}
+ */
+class EngineDecoratorProvider implements IEngineDecoratorProvider {
+  constructor(plugin: IEngineDecoratorProvider) {
```
OrenMe, line 10: this means we are still passing the entire plugin to playkit. lets pass only the decorator provider.

Yuvalke: we're passing it but not keeping it

Yuvalke: Provider will be getName and getEngineDecorator instead of the plugin itself

OrenMe (approved):

PR #545, opened 2021-02-24
fix(FEC-11035): playback error dispatch from player instead of from engine

Yuvalke:

Issue: engine doesn't expose an error for playback error, it dispatched from the player instead - the engine decorator can't ignore the playback error.
Solution: change the dispatch from the player to the engine to make sure the playback error could reject on the engine decorator.
tests failed for scenario _nativeTextTracksMap has index 0 and 2(metadata between) so for raise an error when it gets to i = 1 that doesn't exist. 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -908,10 +908,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     const pkTextTracks = this._getPKTextTracks();
     const pkOffTrack = pkTextTracks.find(track => track.language === 'off');
     const getActiveVidTextTrackIndex = () => {
-      const activeTrackIndex = this._nativeTextTracksMap.findIndex(
-        textTrack => textTrack && this._getDisplayTextTrackModeString() === textTrack.mode
-      );
-      return activeTrackIndex ? activeTrackIndex : -1;
+      return this._nativeTextTracksMap.findIndex(textTrack => textTrack && this._getDisplayTextTrackModeString() === textTrack.mode);
     };
     NativeAdapter._logger.debug('Video element text track change');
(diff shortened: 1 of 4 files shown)


src/player.js
```diff
@@ -1979,10 +1979,8 @@ export default class Player extends FakeEventTarget {
           }
           this._updateTracks(data.tracks);
           this.dispatchEvent(new FakeEvent(CustomEventType.TRACKS_CHANGED, {tracks: this._tracks}));
-          resetFlags();
         })
-        .catch(error => {
-          this.dispatchEvent(new FakeEvent(Html5EventType.ERROR, error));
+        .finally(() => {
```
OrenMe, line 1983: not supported in IE11, please verify we transpile this.
https://caniuse.com/promise-finally

Yuvalke: It's OK, we're using it with post-roll on ads controller as well, tested it as well on IE11

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -899,13 +899,10 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
-        const textTrack = this._nativeTextTracksMap[i];
-        if (this._getDisplayTextTrackModeString() === textTrack.mode) {
-          return i;
-        }
-      }
-      return -1;
+      const activeTrackIndex = this._nativeTextTracksMap.findIndex(
+        textTrack => textTrack && this._getDisplayTextTrackModeString() === textTrack.mode
+      );
+      return activeTrackIndex ? activeTrackIndex : -1;
```
yairans, line 905: findIndex already returns -1 for non existing

yairans (changes requested):

yairans (approved):

PR #546, opened 2021-02-24
fix: decorator API called with incorrect this(proxy) instead of the activeDecorator

Yuvalke:

Issue: API method which overrides by decorator has the proxy as this, it doesn't let us use to register events on the decorator scope and other scope issues.
Solution: make a call directly to decorator API with arguments.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/engine-decorator.js
+++ b/src/engines/engine-decorator.js
@@ -42,8 +42,13 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
           return this._listeners;
         }
         const activeDecorator = this._pluginDecorators.find(decorator => decorator.active);
+        const isGetter = (obj, prop) => !!Object.getOwnPropertyDescriptor(obj, prop)['get'];
         // $FlowFixMe
-        return activeDecorator && prop in activeDecorator ? activeDecorator[prop] : obj[prop];
+        return activeDecorator && prop in activeDecorator
+          ? isGetter(activeDecorator.constructor.prototype, prop)
+            ? activeDecorator[prop]
+            : args => activeDecorator[prop](args)
+          : obj[prop];
       },
       set: (obj, prop, value) => {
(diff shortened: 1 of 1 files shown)


src/engines/engine-decorator.js
```diff
@@ -42,8 +42,16 @@ class EngineDecorator extends FakeEventTarget implements IEngineDecorator {
           return this._listeners;
         }
         const activeDecorator = this._pluginDecorators.find(decorator => decorator.active);
-        // $FlowFixMe
-        return activeDecorator && prop in activeDecorator ? activeDecorator[prop] : obj[prop];
+        const isGetter = (object, property) => {
+          const descriptor = Object.getOwnPropertyDescriptor(object, property);
+          return descriptor && !!descriptor['get'];
+        };
+        const activeEngine = activeDecorator && prop in activeDecorator ? activeDecorator : obj;
```
OrenMe, line 49: sounds like the proper name for this is just `target` as it can be the active decorator or the engine

OrenMe:
```suggestion
        const target = activeDecorator && prop in activeDecorator ? activeDecorator : obj;
```

src/player.js
```diff
@@ -1700,8 +1700,7 @@ export default class Player extends FakeEventTarget {
       this._appendEngineEl();
     } else {
       if (this._engine.id === Engine.id) {
-        // The restoring must be done by the engine itself not by the proxy (engine decorator if exists) to make sure the engine events fired by the engine itself.
-        this._engine.restore.call(this._engine._engine || this._engine, source, this._config);
+        this._engine.restore.call(this._engine, source, this._config);
```
yairans, line 1703:
```suggestion
        this._engine.restore(source, this._config);
```

yairans (changes requested):

yairans (approved):

PR #549, opened 2021-03-03
feat(FEC-10980): Enable setting ABR settings on-the-fly

Yuvalke:

add an API applyABRRestriction to apply ABR restrictions and show only available tracks after restriction.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -1298,7 +1298,7 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _applyABRRestriction(config: Object): void {
-    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine && this._tracks) {
+    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine && this._tracks.length) {
       const {restrictions} = this._config.abr;
       const videoTracks = this._tracks.filter(track => track instanceof VideoTrack);
       const newVideoTracks = filterTracksByRestriction(videoTracks, restrictions);
--- a/src/player.js
+++ b/src/player.js
@@ -1313,10 +1313,15 @@ export default class Player extends FakeEventTarget {
       const newVideoTracks = filterTracksByRestriction(videoTracks, restrictions);
       if (newVideoTracks.length) {
         const currentVideoTracks = this._availableTracks.filter(track => track instanceof VideoTrack);
-        this._engine.applyABRRestriction(restrictions);
-        if (JSON.stringify(currentVideoTracks) !== JSON.stringify(newVideoTracks)) {
+        const videoTracksHasntChanged =
+          currentVideoTracks.length === newVideoTracks.length &&
+          currentVideoTracks.every(function (element, index) {
+            return element.bandwidth === newVideoTracks[index].bandwidth;
+          });
+        if (!videoTracksHasntChanged) {
+          this._engine.applyABRRestriction(restrictions);
           this._availableTracks = this._allTracks.filter(track => !(track instanceof VideoTrack)).concat(newVideoTracks);
-          this.dispatchEvent(new FakeEvent(CustomEventType.TRACKS_CHANGED, {tracks: newVideoTracks}));
+          this.dispatchEvent(new FakeEvent(CustomEventType.TRACKS_CHANGED, {tracks: this._availableTracks}));
         }
       } else {
         Player._logger.warn('nothing has changed values does not meet the restriction');
--- a/src/player.js
+++ b/src/player.js
@@ -2226,6 +2226,7 @@ export default class Player extends FakeEventTarget {
   _updateTracks(tracks: Array<Track>): void {
     Player._logger.debug('Tracks changed', tracks);
     this._availableTracks = this._allTracks = tracks.concat(this._externalCaptionsHandler.getExternalTracks(tracks));
+    this._applyABRRestriction(this._config);
     this._addTextTrackOffOption();
     this._maybeSetTracksLabels();
     this._setDefaultTracks();
--- a/src/player.js
+++ b/src/player.js
@@ -1304,7 +1304,7 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   _applyABRRestriction(config: Object): void {
-    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._allTracks) {
+    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine) {
       const videoTracks = this._allTracks.filter(track => track instanceof VideoTrack);
       const newVideoTracks = this._filterVideoTrackNotInRange(videoTracks);
       const currentVideoTracks = this._availableTracks.filter(track => track instanceof VideoTrack);
@@ -2051,12 +2051,18 @@ export default class Player extends FakeEventTarget {
   _filterVideoTrackNotInRange(tracks: Array<Track>): Array<Track> {
     if (this._config.abr.enabled) {
       if (this._config.abr.restrictions) {
-        let videoTrackInRestriction = tracks.filter(
-          track =>
-            track instanceof VideoTrack &&
-            track.bandwidth >= this._config.abr.restrictions.minBitrate &&
-            track.bandwidth <= this._config.abr.restrictions.maxBitrate
-        );
+        let videoTrackInRestriction = [];
+        const {maxHeight, maxWidth} = this._config.abr.restrictions;
+        if (maxHeight || maxWidth) {
+          videoTrackInRestriction = tracks.filter(
+            track => track instanceof VideoTrack && track.height <= (maxHeight || Infinity) && track.width <= (maxWidth || Infinity)
+          );
+        } else {
+          const {minBitrate, maxBitrate} = this._config.abr.restrictions;
+          videoTrackInRestriction = tracks.filter(
+            track => track instanceof VideoTrack && track.bandwidth >= (minBitrate || 0) && track.bandwidth <= (maxBitrate || Infinity)
+          );
+        }
         const noVideoTrack = tracks.filter(track => !(track instanceof VideoTrack));
         return videoTrackInRestriction.concat(noVideoTrack);
       }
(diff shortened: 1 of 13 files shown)


src/player.js
```diff
@@ -1286,6 +1296,25 @@ export default class Player extends FakeEventTarget {
...
+   */
+  _applyABRRestriction(config: Object): void {
+    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._allTracks) {
+      const videoTracks = this._allTracks.filter(track => track instanceof VideoTrack);
+      const newVideoTracks = this._filterVideoTrackNotInRange(videoTracks);
+      const currentVideoTracks = this._availableTracks.filter(track => track instanceof VideoTrack);
+      this._engine.applyABRRestriction(this._config.abr);
+      if (JSON.stringify(currentVideoTracks) !== JSON.stringify(newVideoTracks)) {
+        this._availableTracks = this._allTracks.filter(track => !(track instanceof VideoTrack)).concat(newVideoTracks);
+        this.dispatchEvent(new FakeEvent(CustomEventType.TRACKS_CHANGED, {tracks: newVideoTracks}));
```
OrenMe, line 1314: Isn't it is better to do this actual filtering in the adapter itself, and get notified about the new tracks.
The engine should know better what tracks are available to it after applying filtering.
I think Shaka does this, while maybe hlsjs doesn't. Maybe good chance to donate this to hlsjs :-)

OrenMe: the bitrate is not the only filterable attribute, we will add also height and width in the future and `_filterVideoTrackNotInRange` will become increasingly more complex.

Yuvalke: I don't think there is reason to use it on engine level if it's relevant for all engines, adapters return the levels from manifest the restriction relevant for their adaptive selection

Yuvalke: Tracks still available if you want to select a track, not from the range - it's possible in Shaka

OrenMe: What I mean is that we assume the internal filtering of each engine will yield same result as this filtering does, and this can lead to incorrect data.

src/utils/restrictions.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/**
+ * Filter the video tracks which not in the range
+ * @function filterVideoTracksByRestriction
+ * @param {Array<Track>} tracks - The tracks to filter
+ * @param {PKABRRestrictionObject} restriction - The restriction
+ * @returns {Array<Track>} - The relevant video tracks after restrictions.
+ */
+function filterTracksByRestriction(tracks: Array<Track>, restriction: PKABRRestrictionObject): Array<VideoTrack> {
```
yairans, line 35: is it's for video tracks only better to get video tracks

src/utils/restrictions.js
```diff
@@ -0,0 +1,41 @@
...
+  const videoTracks = tracks.filter(track => track instanceof VideoTrack);
+  const filterVideoTracks = _filterVideoTracksByRestriction(videoTracks, restriction);
+  return filterVideoTracks.length ? filterVideoTracks : videoTracks;
```
yairans, line 38: why to not return empty array?

test/src/player.spec.js
```diff
@@ -598,29 +598,76 @@ describe('Player', function () {
...
-      player.getTracks().length.should.be.equal(6);
+      player.getTracks().length.should.be.equal(7);
     });
 
     it('should return video tracks', () => {
+      player.getTracks('video').length.should.be.equal(2);
+    });
+
+    it('should return video tracks after restriction include the highest', () => {
+      player.configure({
```
yairans, line 636: please add tests also for height and width

yairans (changes requested):

src/player.js
```diff
@@ -479,6 +487,7 @@ export default class Player extends FakeEventTarget {
       Utils.Object.mergeDeep(this._config, config);
     }
     this._applyTextTrackConfig(config);
+    this._applyABRRestriction(config);
```
yairans, line 490: so there is no api `applyABRRestriction` (as in the PR desc) ? only by configure?

src/player.js
```diff
@@ -1286,6 +1297,25 @@ export default class Player extends FakeEventTarget {
...
+  /**
+   * update the ABR restriction config
+   * @function _applyABRRestriction
+   * @returns {void}
+   * @param {Object} config - new config which configure for checking if it relevant config has changed
+   * @private
+   */
+  _applyABRRestriction(config: Object): void {
+    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine) {
+      const {restrictions} = this._config.abr;
```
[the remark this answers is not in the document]
Yuvalke: No, first check with `config` uses for checking if the current request config has `abr.restrictions`.
this._config used after cause could be the previous config which requests and have to take into account.

src/utils/restrictions.js
Yuvalke: I'm getting all and return the tracks after restriction so it'll return the only video tracks.

src/utils/restrictions.js
Yuvalke: cause if nothing in the range we couldn't play the stream, we avoid it on the adapters.

src/player.js
Yuvalke: Yes, discuss with Oren a long time ago and we decided that it much more make sense to keep the configuration in one code path.

OrenMe (approved):

src/player.js
```diff
@@ -1288,6 +1299,30 @@ export default class Player extends FakeEventTarget {
...
+      const videoTracks = this._allTracks.filter(track => track instanceof VideoTrack);
+      const newVideoTracks = filterTracksByRestriction(videoTracks, restrictions);
+      if (newVideoTracks.length) {
+        const currentVideoTracks = this._availableTracks.filter(track => track instanceof VideoTrack);
+        this._engine.applyABRRestriction(restrictions);
+        if (JSON.stringify(currentVideoTracks) !== JSON.stringify(newVideoTracks)) {
```
yairans, line 1317: can we do it in better performance? just run on the objects and compare by bandwidth

src/player.js
```diff
@@ -1288,6 +1299,30 @@ export default class Player extends FakeEventTarget {
...
+          this._availableTracks = this._allTracks.filter(track => !(track instanceof VideoTrack)).concat(newVideoTracks);
+          this.dispatchEvent(new FakeEvent(CustomEventType.TRACKS_CHANGED, {tracks: newVideoTracks}));
+        }
+      } else {
+        Player._logger.warn('nothing has changed values does not meet the restriction');
```
yairans, line 1322:
```suggestion
        Player._logger.warn('Nothing has changed values do not meet the restriction');
```
I'm not sure `values` is clear term here

src/player.js
```diff
@@ -1288,6 +1299,30 @@ export default class Player extends FakeEventTarget {
...
+   * @private
+   */
+  _applyABRRestriction(config: Object): void {
+    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine) {
+      const {restrictions} = this._config.abr;
+      const videoTracks = this._allTracks.filter(track => track instanceof VideoTrack);
+      const newVideoTracks = filterTracksByRestriction(videoTracks, restrictions);
+      if (newVideoTracks.length) {
+        const currentVideoTracks = this._availableTracks.filter(track => track instanceof VideoTrack);
+        this._engine.applyABRRestriction(restrictions);
```
yairans, line 1316: this line should be in the if below no?
also `this._engine` could be null this moment

yairans (changes requested):

src/player.js
```diff
@@ -1288,6 +1299,30 @@ export default class Player extends FakeEventTarget {
...
 
+  /**
+   * update the ABR restriction config
+   * @function _applyABRRestriction
+   * @returns {void}
+   * @param {Object} config - new config which configure for checking if it relevant config has changed
+   * @private
+   */
+  _applyABRRestriction(config: Object): void {
+    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine) {
```
yairans, line 1310: actually, this is no only ABR settings but also related to the video tracks options displayed to the user

yairans (changes requested):

src/player.js
```diff
@@ -1288,6 +1299,30 @@ export default class Player extends FakeEventTarget {
...
+      const {restrictions} = this._config.abr;
+      const videoTracks = this._allTracks.filter(track => track instanceof VideoTrack);
+      const newVideoTracks = filterTracksByRestriction(videoTracks, restrictions);
+      if (newVideoTracks.length) {
+        const currentVideoTracks = this._availableTracks.filter(track => track instanceof VideoTrack);
+        this._engine.applyABRRestriction(restrictions);
+        if (JSON.stringify(currentVideoTracks) !== JSON.stringify(newVideoTracks)) {
+          this._availableTracks = this._allTracks.filter(track => !(track instanceof VideoTrack)).concat(newVideoTracks);
+          this.dispatchEvent(new FakeEvent(CustomEventType.TRACKS_CHANGED, {tracks: newVideoTracks}));
```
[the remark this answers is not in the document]
Yuvalke: 2. if there's no change it won't dispatch the `TRACKS_CHANGED`. the adapter should dispatch `TRACKS_CHANGED` with all the possible tracks filtering process happens on player level. It'll be 2 events of `TRACKS_CHANGED`

src/player.js
Yuvalke: engine checked previously

src/player.js
Yuvalke: Performence wise it's better then loop over one of them

yairans (changes requested):

src/player.js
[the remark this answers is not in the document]
yairans: do you mean before load this does nothing?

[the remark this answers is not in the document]
Yuvalke: it's not before load

src/player.js
Yuvalke: much less

src/player.js
```diff
@@ -1288,6 +1290,43 @@ export default class Player extends FakeEventTarget {
...
 
+  /**
+   * update the ABR restriction config
+   * @function _applyABRRestriction
+   * @returns {void}
+   * @param {Object} config - new config which configure for checking if it relevant config has changed
+   * @private
+   */
+  _applyABRRestriction(config: Object): void {
+    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine && this._tracks) {
```
yairans, line 1301:
```suggestion
    if (Utils.Object.hasPropertyPath(config, 'abr.restrictions') && this._engine && this._tracks.length) {
```

yairans (changes requested):

yairans (approved):

PR #584, opened 2021-07-07
feat(FEC-10988): refactor 608/708 default player handling

Yuvalke:

Issue: safari shows a placeholder for 708 captions even if no track available.
Solutions: set all captions hidden in UI and change one of them to hidden, this one will indicate if cues are coming or not.
when cues are coming we're showing all the captions.

### CheckLists

- [ ] Docs have been updated


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -686,8 +686,12 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
   _getParsedTextTracks(): Array<PKTextTrack> {
     const captionsTextTrackLabels = [this._config.captionsTextTrack1Label, this._config.captionsTextTrack2Label];
     const captionsTextTrackLanguageCodes = [this._config.captionsTextTrack1LanguageCode, this._config.captionsTextTrack2LanguageCode];
+    const captionAvailable = !!Array.from(this._videoElement.textTracks).find(
+      track => track.kind === 'captions' && ((track.activeCues && track.activeCues.length > 0) || track.mode === 'showing')
```
OrenMe, line 690: it should be cues in general and not just activeCues
Why do we need showing state? maybe because you do not check just cues? I think it is wrong cause native video player may have this option set and then mode will be `showing`

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -704,15 +708,29 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           } else if (settings.kind === 'captions' && this._config.enableCEA708Captions) {
             settings.label = settings.label || captionsTextTrackLabels.shift();
             settings.language = settings.language || captionsTextTrackLanguageCodes.shift();
+            settings.available = captionAvailable;
```
OrenMe, line 711: I missed this in the ABR PR, this `available` attribute looks very funky to me but I understand it serves us here
IMO we shouldn't extend the external object we expose.
I would have kept the internal full track object and expose an external one which is filtered with the only available tracks instead of exposing it to all with this flag.

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -704,15 +708,29 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
       }
     }
     return parsedTracks;
   }
 
+  _maybeShow708Captions(textTrack: any): void {
+    textTrack.mode = 'hidden';
+    this._eventManager.listenOnce(textTrack, 'cuechange', () => {
+      const textTracks = this._getPKTextTracks();
+      textTracks.forEach(track => (track.available = true) && (track.mode = 'disable'));
```
OrenMe, line 730: if we have the language of the track as user selected will we still auto enable it?

OrenMe (changes requested):

src/utils/util.js
```diff
@@ -638,4 +639,53 @@ const _VERSION = {
...
+  },
+
+  /**
+   * if text track is subtitle or caption
+   * @public
+   * @param {any} track  - text track
+   * @returns {void}
+   */
+  isNativeTextTrack: function (track: any) {
+    return track && [this.SUBTITLES, this.CAPTIONS].includes(track.kind);
```
OrenMe, line 668: maybe safer to use _textTrack.SUBTITLES etc...

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -690,29 +699,48 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
             settings.label = settings.label || captionsTextTrackLabels.shift();
             settings.language = settings.language || captionsTextTrackLanguageCodes.shift();
+            settings.available = this._captionsHidden;
             parsedTracks.push(new PKTextTrack(settings));
             this._nativeTextTracksMap[settings.index] = textTracks[i];
           }
         }
       }
+
+      if (!this._captionsHidden) {
```
OrenMe, line 723: according to this `_captionsHidden` will always be hidden at first, but we are not checking if maybe the track has cues already which will mean it is already available

OrenMe (changes requested):

src/engines/html5/media-source/adapters/native-adapter.js
Yuvalke: Correct, there's no use case we're starting with cues or activeCues.
textTrack is `disabled` cues and activeCues removed from text track.
so it'll always start in `disabled` or `showing` because safari auto-selection but won't load any cue yet(right after load).
But there's a use case that new textTrack add, then we have an indicator for if caption still hidden

src/utils/util.js
```diff
@@ -638,4 +639,53 @@ const _VERSION = {
...
 
-export {_Number as Number, _String as String, _Object as Object, _Generator as Generator, _Dom as Dom, _Http as Http, _VERSION as VERSION};
+const _textTrack = {
+  DISABLED: 'disabled',
+  SHOWING: 'showing',
+  HIDDEN: 'hidden',
+
+  METADATA: 'metadata',
+  SUBTITLES: 'subtitles',
+  CAPTIONS: 'captions',
```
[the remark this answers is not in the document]
Yuvalke: it's true and also the helper methods.

OrenMe (dismissed):

PR #587, opened 2021-07-09
fix(FEC-11050): some custom events are not dispatched from the player

JonathanTGold:

issue: Some events which are in the custom events enum are not listened by, and dispatched outside the player.
fix: listen and dispatch all these events.FEC-11050

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -305,7 +305,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     this._eventManager.listen(this._el, Html5EventType.ERROR, () => this._handleVideoError());
     this._handleMetadataTrackEvents();
     this._eventManager.listen(this._el.textTracks, 'addtrack', (event: any) => {
-      if (event.track.kind === 'captions' || event.track.kind === 'subtitles') {
+      if (PKTextTrack.isNativeTextTrack(event.track)) {
         this.dispatchEvent(new FakeEvent(CustomEventType.TEXT_TRACK_ADDED, {track: event.track}));
       }
     });
@@ -1068,7 +1068,7 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @private
    */
   _addCueChangeListener(): void {
-    let textTrackEl = Array.from(this._el.textTracks).find(track => track && track.mode !== 'disabled');
+    let textTrackEl = Array.from(this._el.textTracks).find(track => PKTextTrack.isNativeTextTrack(track) && track.mode !== PKTextTrack.MODE.DISABLED);
     if (textTrackEl) {
       this._eventManager.listen(textTrackEl, 'cuechange', (e: FakeEvent) => this._onCueChange(e));
     }
@@ -1114,7 +1114,9 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @returns {void}
    */
   resetAllCues(): void {
-    let activeTextTrack = Array.from(this._el.textTracks).find(track => track && track.mode !== 'disabled');
+    let activeTextTrack = Array.from(this._el.textTracks).find(
+      track => PKTextTrack.isNativeTextTrack(track) && track.mode !== PKTextTrack.MODE.DISABLED
+    );
     if (activeTextTrack) {
       for (let i = 0; i < activeTextTrack.cues.length; i++) {
         activeTextTrack.cues[i].hasBeenReset = true;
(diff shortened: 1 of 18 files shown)


src/player.js
```diff
@@ -1838,7 +1841,10 @@ export default class Player extends FakeEventTarget {
       });
       this._eventManager.listen(this._engine, CustomEventType.TEXT_TRACK_CHANGED, (event: FakeEvent) => this._onTextTrackChanged(event));
       this._eventManager.listen(this._engine, CustomEventType.TRACKS_CHANGED, (event: FakeEvent) => this._onTracksChanged(event));
-      this._eventManager.listen(this._engine, CustomEventType.TEXT_CUE_CHANGED, (event: FakeEvent) => this._onCueChange(event));
+      this._eventManager.listen(this._engine, CustomEventType.TEXT_CUE_CHANGED, (event: FakeEvent) => {
+        this._onCueChange(event);
+        this.dispatchEvent(event);
```
yairans, line 1846: `TEXT_CUE_CHANGED` triggered by the external caption handler so need to consider it as well.
better to dispatch to from `_onCueChange` so it will be triggered for both.

yairans (changes requested):

src/player.js
JonathanTGold: i added in the engine

JonathanTGold: I moved it there, but I'm not sure about the timing, whether it should be triggered after the  _onCueChange logic as in _onTextTrackChanged or before (immediately upon the entering the function )

yairans (dismissed):

src/player.js
```diff
@@ -1272,6 +1272,9 @@ export default class Player extends FakeEventTarget {
         } else if (track.external) {
           this._engine.hideTextTrack();
           this._externalCaptionsHandler.selectTextTrack(track);
+          if (this.config.text.useNativeTextTrack) {
+            this._engine.addCueChangeListener();
+          }
```
OrenMe, line 1277: this feels like we have some issue we need to align internally as we don't want to expose this outside of engine as API(too much implementation details leaking on API surface).
How did this work until now? is this only to expose the cue change event?

JonathanTGold: It's just to expose the cue_change event, 
in the present flow it's the only choice to listen to this event (in the specific case of an externals subtitles combined  with using native TextTreack) 
since the component which  produces this event is the engine (and that by listening to the native cuechange event) in player instruction each time engine.selectTrack is called, 
And what happens is that in the case of externals, it does not call the selectTrack of the engine, but mange it through the externalCaptionsHandler which it interact directly withe the HTML

but maybe it would be a better solution listening directly to the native cuechange event in the extranlCaptionsHandler just like the engine (because in this case it really replaces the engine's place when it comes to subtitles.

src/track/external-captions-handler.js
```diff
@@ -231,6 +231,15 @@ class ExternalCaptionsHandler extends FakeEventTarget {
     }
   }
 
+  addCueChangeListener(textTrackEl: TextTrack): void {
+    this._eventManager.listen(textTrackEl, 'cuechange', (e: FakeEvent) => this._onCueChange(e));
```
yairans, line 235: need to add unlisten ass well. see the engine when is needed

src/track/external-captions-handler.js
```diff
@@ -231,6 +231,15 @@ class ExternalCaptionsHandler extends FakeEventTarget {
...
+  }
+
+  _onCueChange(e: FakeEvent): void {
```
yairans, line 238: pls see is if you can use this method also here 
https://github.com/kaltura/playkit-js/blob/v0.72.0/src/track/external-captions-handler.js#L382
anyway need to verify that the active cues there are normalized

yairans (changes requested):

src/track/external-captions-handler.js
JonathanTGold: the two method called in two different cases, the other method you talk about called only in case of using costume caption (look at lines 216/219)

src/track/external-captions-handler.js
JonathanTGold: added

src/track/external-captions-handler.js
```diff
@@ -231,6 +233,38 @@ class ExternalCaptionsHandler extends FakeEventTarget {
...
+    this._eventManager.listen(textTrackEl, 'cuechange', (e: FakeEvent) => this._onCueChange(e));
+  }
+
+  /**
+   * Remove cuechange listeners from textTracks
+   * @returns {void}
+   * @private
+   */
+  _removeCueChangeListeners(): void {
+    for (let i = 0; i < this._player.getVideoElement().textTracks.length; i++) {
```
yairans, line 252: need to verify `this._player.getVideoElement()` and `textTracks` existence
like we do in `_addCuesToNativeTextTrack`

yairans (changes requested):

src/track/external-captions-handler.js
JonathanTGold: done

src/track/external-captions-handler.js
```diff
@@ -231,6 +233,42 @@ class ExternalCaptionsHandler extends FakeEventTarget {
     }
   }
 
+  /**
+   * Add cuechange listener to active textTrack.
+   * @param {TextTrack} textTrackEl - selected text track
+   * @returns {void}
+   * @private
+   */
+  addCueChangeListener(textTrackEl: TextTrack): void {
```
yairans, line 242: where do you call this method?

JonathanTGold: fixed

src/track/external-captions-handler.js
```diff
@@ -209,10 +211,13 @@ class ExternalCaptionsHandler extends FakeEventTarget {
       }
     }
   }
+
   _selectTextTrack(textTrack: TextTrack) {
     this.hideTextTrack();
     if (this._player.config.text.useNativeTextTrack) {
       this._addCuesToNativeTextTrack(this._textTrackModel[textTrack.language].cues);
+      this._removeCueChangeListeners();
```
yairans, line 219: already removed by `hideTextTrack` no?

src/engines/html5/html5.js
```diff
@@ -6,15 +6,15 @@ import {CustomEventType, Html5EventType} from '../../event/event-type';
...
+import PKTextTrack, {getActiveCues} from '../../track/text-track';
 import ImageTrack from '../../track/image-track';
-import {Cue} from '../../track/vtt-cue';
 import * as Utils from '../../utils/util';
 import Html5AutoPlayCapability from './capabilities/html5-autoplay';
 import Error from '../../error/error';
 import getLogger from '../../utils/logger';
 import {DroppedFramesWatcher} from '../dropped-frames-watcher';
 import {ThumbnailInfo} from '../../thumbnail/thumbnail-info';
+import TextTrack from '../../track/text-track';
```
yairans, line 17: already imported in line 9 as PKTextTrack. where it used?

src/track/external-captions-handler.js
```diff
@@ -11,6 +11,7 @@ import EventManager from '../event/event-manager';
 import FakeEventTarget from '../event/fake-event-target';
 import {Cue} from './vtt-cue';
 import Player from '../player';
+import PKTextTrack from './text-track';
```
yairans, line 14: already imported in line 5 as TextTrack

yairans (changes requested):

yairans (approved):

PR #588, opened 2021-07-14
fix(FEC-11402): preview doesn't work in video ratio different than 16:9

RoyBregman:

To calculate the preview image in KP we need to get the videoHeight and videoWidth so I exposed them in the core player

needed for https://github.com/kaltura/kaltura-player-js/pull/466
Solves FEC-11402

### CheckLists

- [X] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -595,6 +595,30 @@ export default class Player extends FakeEventTarget {
   }
 
   /**
+   * Get video height.
+   * @returns {?number} - The the intrinsic height of the video.
+   * @public
+   */
+  get videoHeight(): ?number {
+    if (this._engine) {
+      return this._engine.videoHeight;
+    }
+    return null;
+  }
+
+  /**
+   * Get video width.
+   * @returns {?number} - The the intrinsic width of the video.
+   * @public
+   */
+  get videoWidth(): ?number {
+    if (this._engine) {
+      return this._engine.videoWidth;
+    }
+    return null;
+  }
+
+  /**
    * Resets the necessary components before change media.
    * @public
    * @returns {void}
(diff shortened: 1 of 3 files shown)


src/player.js
```diff
@@ -594,6 +594,30 @@ export default class Player extends FakeEventTarget {
...
+  /**
+   * Get video height.
+   * @returns {?number} - The intrinsic height of the video.
+   * @public
+   */
+  get videoHeight(): ?number {
+    if (this._engine) {
+      return this._engine.videoHeight;
+    }
+    return null;
```
yairans, line 606: I would return 0 so the type can be number

src/player.js
```diff
@@ -594,6 +594,30 @@ export default class Player extends FakeEventTarget {
...
+  /**
+   * Get video width.
+   * @returns {?number} - The intrinsic width of the video.
+   * @public
+   */
+  get videoWidth(): ?number {
+    if (this._engine) {
+      return this._engine.videoWidth;
+    }
+    return null;
```
yairans, line 618: same

src/player.js
RoyBregman: I thought undefined was more appropriate actually. But the reason I chose to return null is to be aligned with all other number getters we have in the player. For example - 
https://github.com/kaltura/playkit-js/blob/ee7653f7cae10fcf2e18e581e8661ccf9308173f/src/player.js#L816
https://github.com/kaltura/playkit-js/blob/ee7653f7cae10fcf2e18e581e8661ccf9308173f/src/player.js#L828
https://github.com/kaltura/playkit-js/blob/ee7653f7cae10fcf2e18e581e8661ccf9308173f/src/player.js#L841

yairans (approved):

PR #593, opened 2021-08-03
fix(FEC-11451): video tracks are always returned with active false in…

JonathanTGold:

**issue**: video tracks are always returned with active false in native adapter, so when you call kalturaPlayer.getActiveTracks() you always get the video track undefined because it is filtered by active = true

**fix**: add the missing index property in the videoTrack object, 
look at this line https://github.com/kaltura/playkit-js/blob/master/src/player.js#L2381,
Which is the cause of the problem, since the index property does not exist, the condition will always return false

**solves: (FEC-11451)**

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -545,7 +545,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           height: videoHeight,
           width: videoWidth,
           active: true,
-          index: 0
+          index: Array.from(videoTracks).findIndex((track: VideoTrack) => track.selected)
         };
         this._onTrackChanged(new VideoTrack(setting));
       }
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -544,7 +544,8 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           language: '',
           height: videoHeight,
           width: videoWidth,
-          active: true
+          active: true,
+          index: 0
         };
         this._onTrackChanged(new VideoTrack(setting));
       }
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -643,13 +644,13 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
   _getParsedAdaptiveVideoTracks(): Array<Track> {
-    //TODO check adaptation in safari hls
     const videoTracks = this._videoElement.videoTracks;
     const parsedTracks = [];
     if (videoTracks) {
       for (let i = 0; i < videoTracks.length; i++) {
         const settings = {
-          //TODO calculate width/height/bandwidth
+          height: this._videoElement.videoHeight,
+          width: this._videoElement.videoWidth,
```
OrenMe, line 653: this value will change on change of video track so we can't know it upfront like this AFAIK

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -544,7 +544,8 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           language: '',
           height: videoHeight,
           width: videoWidth,
-          active: true
+          active: true,
+          index: 0
```
OrenMe, line 548: this means we will always say the index is 0 even if track changes

OrenMe (dismissed):

yairans (dismissed):

yairans:
@JonathanTGold is there a test you can update with the new index?

yairans (approved):

src/engines/html5/media-source/adapters/native-adapter.js
JonathanTGold: fixed

src/engines/html5/media-source/adapters/native-adapter.js
JonathanTGold: fixed

PR #605, opened 2021-09-14
fix(FEC-11537): Need to re-add fontScale to textStyle config for backward compatability

SivanA-Kaltura:

Re-add fontScale to textStyle for backward compatibility.
Fixes FEC-11537.

### CheckLists

- [x] Docs have been updated

### DIFF
--- a/src/track/text-style.js
+++ b/src/track/text-style.js
@@ -148,6 +148,7 @@ class TextStyle {
     let textStyle = new TextStyle();
     textStyle.fontEdge = getValue(setting.fontEdge, textStyle.fontEdge);
     textStyle.fontSize = getValue(setting.fontSize, textStyle.fontSize);
+    textStyle.fontScale = getValue(setting.fontScale, textStyle.fontScale);
     textStyle.fontColor = getValue(setting.fontColor, textStyle.fontColor);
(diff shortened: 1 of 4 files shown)


src/track/text-style.js
```diff
@@ -148,6 +148,7 @@ class TextStyle {
     let textStyle = new TextStyle();
     textStyle.fontEdge = getValue(setting.fontEdge, textStyle.fontEdge);
     textStyle.fontSize = getValue(setting.fontSize, textStyle.fontSize);
+    textStyle.fontScale = getValue(setting.fontScale, textStyle.fontScale);
```
yairans, line 151: so if app overrides the `fontSize` only, the `fontScale` wiill override it and it won't be changed no?

SivanA-Kaltura: In that case, getValue will receive settings.fontScale=undefined, and will fallback to textStyle.fontScale (which was set by fontSize)

yairans: great. do we have unit tests for both fontSize and fontScale overrriding?

SivanA-Kaltura: Yes, see my changes in player.spec.js

SivanA-Kaltura: Yes, see my changes to player.spec.js

yairans (approved):

PR #607, opened 2021-09-19
fix(FEC-11547): The thumbnalils images should be relative to the VTT file adress by default

JonathanTGold:

Currently the thumbnalils images URL is relative to the application domain by default, Although the convention is that they should be relative to the vtt file adress

solves: FEC-11547

related pr: https://github.com/kaltura/kaltura-player-js/pull/494

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/flow-typed/types/exteranl-thumbnails-object.js
+++ b/flow-typed/types/exteranl-thumbnails-object.js
@@ -1,5 +1,4 @@
 // @flow
 declare type PKExternalThumbnailsConfig = {
-  imgBaseUrl?: string,
   vttUrl: string
 };
--- a/src/thumbnail/external-thumbnails-handler.js
+++ b/src/thumbnail/external-thumbnails-handler.js
@@ -218,7 +218,7 @@ class ExternalThumbnailsHandler extends FakeEventTarget {
    */
   _extractCueMetadata(vttCue: VTTCue, thumbnailsConfig: PKExternalThumbnailsConfig): PKThumbnailVttCue {
     const {startTime, endTime, text} = vttCue;
-    const {imgBaseUrl} = thumbnailsConfig;
+    const imgBaseUrl = thumbnailsConfig.vttUrl.substring(0, thumbnailsConfig.vttUrl.lastIndexOf('/'));
     const isVTTIncludesImgSizeOnly: boolean = VTT_INCLUDES_SIZE_ONLY.test(text);
     const isVTTIncludesImgSizeAndCoords: boolean = VTT_INCLUDES_SIZE_AND_COORDS.test(text);
     let isValidThumbnailVTTFormat: boolean = false;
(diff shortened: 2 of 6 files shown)


flow-typed/types/exteranl-thumbnails-object.js
```diff
@@ -1,5 +1,4 @@
 // @flow
 declare type PKExternalThumbnailsConfig = {
-  imgBaseUrl?: string,
```
yairans, line 3: need to update the docs

src/thumbnail/external-thumbnails-handler.js
```diff
@@ -241,10 +241,10 @@ class ExternalThumbnailsHandler extends FakeEventTarget {
       size = {width, height};
       isValidThumbnailVTTFormat = [x, y, width, height, imgUrl].every(option => option !== undefined);
     } else {
-      imgUrl = text;
+      imgUrl = text[0] === '/' ? text.substring(1) : text;
```
yairans, line 244: I think it always should be relative. unless the text is starting with 'http' or '//'. @OrenMe wdyt?

OrenMe: so if we have a relative url in the vtt file we set the image url as `/myImagePath/1.png` and then it get the hosting page domain instead of the vtt file?
I think that what you say @yairans is more safe, check if thumb image url starts with http|s and if so take it as is, otherwise treat it as relative

JonathanTGold: no, it will always get the vtt file domain (and path) unless it is absolute path

yairans: @JonathanTGold now you consider it as relative only if it's starting with '/'. I think you should consider it relative even it's not starting with '/', provided it's not starting with 'http' or '//'. e.g.
'/thumbnail.png' = 'thumbnail.png' = relative. 
'http//:...thumbnail.png' = '//....thumbnail.png' = absolute.

JonathanTGold: updated, now any case beside '/' in the begining of the string (absulut path) would remain as it is, otherwise would be resolved to relative pathe reating to the vtt file location

OrenMe (dismissed):

src/thumbnail/external-thumbnails-handler.js
```diff
@@ -12,6 +12,9 @@ import {Html5EventType} from '../event/event-type';
 const VTT_INCLUDES_SIZE_ONLY: RegExp = /#wh=/i;
 const VTT_INCLUDES_SIZE_AND_COORDS: RegExp = /#xywh=/i;
 
+const ABSOLUTE_PATH_PATTERN: RegExp = new RegExp('^http.+');
```
OrenMe, line 15:
```suggestion
const ABSOLUTE_PATH_PATTERN: RegExp = new RegExp(/^https?:\/\//);;
```

OrenMe (dismissed):

OrenMe (approved):

PR #620, opened 2021-12-21
fix(FEC-11745): Player_Web_Mac_Safari - Flash of live stream after Start Over

JonathanTGold:

**issue:** in live stream played on safari when you seek back to - 0 (start over), the currentTime jump to live edge (and equal to duration) for a few ms before resets back to 0, making the isOnLiveEdge() logic on the engine to evalute to false
**fix:** move the isOnLiveEdge() api defenition to the adapters level and let the native adapter to overide the logic and return false by default as long as currentTime resetting was not completed

solves: FEC-11745

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -120,6 +120,8 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
 
   _waitingEventTriggered: ?boolean = false;
 
+  _wasCurrentTimeSetSuccessfully: boolean;
+
   _segmentDuration: number = 0;
 
   /**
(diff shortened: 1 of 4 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -1169,6 +1173,10 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     return this._videoElement.duration === Infinity;
   }
 
+  isOnLiveEdge(): boolean {
+    return this._wasCurrentTimeSetSuccessfully ? super.isOnLiveEdge() : false;
```
yairans, line 1177:
```suggestion
    return this._wasCurrentTimeSetSuccessfully  && super.isOnLiveEdge();
```

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -324,6 +326,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
    * @returns {Promise<Object>} - The loaded data
    */
   load(startTime: ?number): Promise<Object> {
+    this._wasCurrentTimeSetSuccessfully = false;
```
yairans, line 329: this is only when startTime exist and is greater than -1

yairans: so you can move it to the else of _setStartTime

yairans (changes requested):
please update the PR desc

yairans (approved):

PR #633, opened 2022-02-10
fix(FEC-11697): 2 captions are selected in the menu

JonathanTGold:

**issue:** the create _addTextTrackOffOption method of the player creates duplicated indexes every time duplicated languages are detected (and warning created);
**fix:** right now the index property of TextTrack is used as a unique identifier but in the other hand we don't mange it as such (the index is created in different places), so , the index management all moved to one place under the TextTrack class

related pr:

https://github.com/kaltura/playkit-js-dash/pull/182
https://github.com/kaltura/playkit-js-hls/pull/161

solves: FEC-11697  & FEC-11893 & FEC-11973

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -976,15 +976,15 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
   _onNativeTextTrackChange(): void {
     const pkTextTracks = this._getPKTextTracks();
     const pkOffTrack = pkTextTracks.find(track => track.language === 'off');
-    const getActiveVidTextTrackIndex = () => {
+    const getActiveVidTextTrackIndex = (): number => {
       for (const textTrackId in this._nativeTextTracksMap) {
         if (this._getDisplayTextTrackModeString() === this._nativeTextTracksMap[textTrackId].mode) {
-          return textTrackId;
+          return Number(textTrackId);
         }
       }
       return -1;
     };
-    const vidIndex = getActiveVidTextTrackIndex();
+    const vidIndex: number = getActiveVidTextTrackIndex();
     const activePKtextTrack = this._getActivePKTextTrack();
     const pkIndex = activePKtextTrack ? activePKtextTrack.index : -1;
     if (vidIndex !== pkIndex) {
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -721,7 +721,6 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     const captionsTextTrackLanguageCodes = [this._config.captionsTextTrack1LanguageCode, this._config.captionsTextTrack2LanguageCode];
     const textTracks = this._videoElement.textTracks;
     const parsedTracks = [];
-    let internalTrackIndex = 0;
     if (textTracks) {
       for (let i = 0; i < textTracks.length; i++) {
         if (!PKTextTrack.isExternalTrack(textTracks[i])) {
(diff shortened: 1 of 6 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -977,7 +977,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     const pkTextTracks = this._getPKTextTracks();
     const pkOffTrack = pkTextTracks.find(track => track.language === 'off');
     const getActiveVidTextTrackIndex = () => {
-      return this._nativeTextTracksMap.findIndex(textTrack => textTrack && this._getDisplayTextTrackModeString() === textTrack.mode);
+      return Object.values(this._nativeTextTracksMap).findIndex(textTrack => textTrack && this._getDisplayTextTrackModeString() === textTrack.mode);
```
yairans, line 980: use the index/key from the map instead of the index in the array which is not relevant

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -977,7 +977,10 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     const pkTextTracks = this._getPKTextTracks();
     const pkOffTrack = pkTextTracks.find(track => track.language === 'off');
     const getActiveVidTextTrackIndex = () => {
-      return this._nativeTextTracksMap.findIndex(textTrack => textTrack && this._getDisplayTextTrackModeString() === textTrack.mode);
+      const activeTrack: TextTrack = Object.entries(this._nativeTextTracksMap).find(
+        ([, textTrack]) => this._getDisplayTextTrackModeString() === textTrack.mode
+      );
+      return activeTrack ? activeTrack.index : -1;
```
yairans, line 983: there is no index on a native text track. you should use the index (key) from the map

yairans (changes requested):

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -977,7 +977,12 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     const pkTextTracks = this._getPKTextTracks();
     const pkOffTrack = pkTextTracks.find(track => track.language === 'off');
     const getActiveVidTextTrackIndex = () => {
-      return this._nativeTextTracksMap.findIndex(textTrack => textTrack && this._getDisplayTextTrackModeString() === textTrack.mode);
+      for (const textTrackId in this._nativeTextTracksMap) {
```
yairans, line 980: can you explain what wrong in the previous impl?

JonathanTGold: the Array.find returns the entire object and then i need to extract the key unlike **for in** that run on the object keys and returns only what i needed from the first place, So I thought it would be more correct and readable to use it
But it is clear that both approaches will do the job

src/track/text-track.js
```diff
@@ -60,6 +91,8 @@ const TextTrack: TextTrack = class TextTrack extends Track {
     this._label = this.label || this.language;
     this._kind = settings.kind;
     this._external = settings.external;
+    this._index = TextTrack._generateIndex();
+    console.error('121212', this);
```
yairans, line 95: remove

src/track/text-track.js
```diff
@@ -41,6 +63,15 @@ const TextTrack: TextTrack = class TextTrack extends Track {
     return this._kind;
   }
 
+  /**
+   * Setter for the kind of the text track.
+   * @public
+   * @param {string} value - The kind of text the track.
+   */
+  set kind(value: string) {
```
yairans, line 71: why it needed?

src/track/text-track.js
```diff
@@ -16,6 +16,28 @@ const TextTrack: TextTrack = class TextTrack extends Track {
...
+   * @returns {number} - the next track index.
+   */
+  static _generateIndex(): number {
+    return TextTrack._tracksCount++;
+  }
+  /**
+   * reset the track count.
+   * @returns {void}
+   */
+  static resetTracksCount(): void {
```
yairans, line 37: I would call it just reset. the track count is an internal stuff

yairans (approved):

PR #634, opened 2022-02-10
feat(FEC-11894): Persist user selection during playlist playback

yairans:

store the video track selection and select it on load 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -324,6 +324,7 @@ export default class Player extends FakeEventTarget {
     muted: undefined,
     volume: undefined,
     rate: undefined,
+    videoHeight: undefined,
     audioLanguage: '',
     textLanguage: ''
   };
@@ -1303,6 +1304,7 @@ export default class Player extends FakeEventTarget {
   selectTrack(track: ?Track): void {
     if (this._engine) {
       if (track instanceof VideoTrack) {
+        this._playbackAttributesState.videoHeight = track.height;
         if (this._stateManager.currentState.type === StateType.IDLE) {
           this._pendingSelectedVideoTrack = track;
         } else {
@@ -1387,6 +1389,7 @@ export default class Player extends FakeEventTarget {
     if (this._engine) {
       this._engine.enableAdaptiveBitrate();
     }
+    this._playbackAttributesState.videoHeight = undefined;
   }
 
   /**
@@ -2549,6 +2552,7 @@ export default class Player extends FakeEventTarget {
     let currentOrConfiguredAudioLang = this._playbackAttributesState.audioLanguage || playbackConfig.audioLanguage;
     this._setDefaultTrack<TextTrack>(this._getTextTracks(), currentOrConfiguredTextLang, offTextTrack);
     this._setDefaultTrack<AudioTrack>(this._getAudioTracks(), currentOrConfiguredAudioLang, activeTracks.audio);
+    this._setDefaultVideoTrack();
   }
 
   /**
@@ -2598,6 +2602,20 @@ export default class Player extends FakeEventTarget {
       } else if (defaultTrack && !defaultTrack.active) {
         this.selectTrack(defaultTrack);
       }
+    }
+  }
+
+  /**
+   * Sets the video track selected by the user.
+   * @returns {void}
+   * @private
+   */
+  _setDefaultVideoTrack() {
+    const selectedVideoTrack = this._getVideoTracks()
+      .sort((track1: VideoTrack, track2: VideoTrack) => track2.bandwidth - track1.bandwidth)
+      .find((track: VideoTrack) => track.height === this._playbackAttributesState.videoHeight);
+    if (selectedVideoTrack) {
+      this.selectTrack(selectedVideoTrack);
     }
   }
 
--- a/src/drm/fairplay.js
+++ b/src/drm/fairplay.js
@@ -24,8 +24,9 @@ const FairPlay: IDrmProtocol = class FairPlay {
    * @return {boolean} - Whether FairPlay can be play on the current environment.
    */
   static canPlayDrm(drmData: Array<PKDrmDataObject>): boolean {
-    _logger.debug('Can play DRM scheme of: ' + DrmScheme.FAIRPLAY);
-    return !!drmData.find(drmEntry => drmEntry.scheme === DrmScheme.FAIRPLAY) && !!window.WebKitMediaKeys;
+    const result = !!drmData.find(drmEntry => drmEntry.scheme === DrmScheme.FAIRPLAY) && !!window.WebKitMediaKeys;
+    _logger.debug(`Can play DRM scheme of: ${DrmScheme.FAIRPLAY} is ${result}`);
+    return result;
   }
 
   /**
--- a/src/drm/fairplay.js
+++ b/src/drm/fairplay.js
@@ -24,7 +24,7 @@ const FairPlay: IDrmProtocol = class FairPlay {
    * @return {boolean} - Whether FairPlay can be play on the current environment.
    */
   static canPlayDrm(drmData: Array<PKDrmDataObject>): boolean {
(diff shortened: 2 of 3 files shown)


src/player.js
```diff
@@ -2601,6 +2605,20 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  /**
+   * Sets the video track selected by the user.
+   * @returns {void}
+   * @private
+   */
+  _setDefaultVideoTrack() {
+    const selectedVideoTrack = this._getVideoTracks()
```
OrenMe, line 2614: is this for ABR or only manual selected bitrate?
If user selected the video track then this will be OK, but if this happens on every loading of media it means this will apply for ABR mode as well.
For ABR mode we should use the estimated bitrate(which for HLS doesn't work well)

src/player.js
```diff
@@ -2601,6 +2605,20 @@ export default class Player extends FakeEventTarget {
...
+      .sort((track1: VideoTrack, track2: VideoTrack) => track2.bandwidth - track1.bandwidth)
+      .find((track: VideoTrack) => track.height && track.height === this._playbackAttributesState.videoHeight);
```
SivanA-Kaltura, line 2616: what about comparing width ? or label ?

src/player.js
```diff
@@ -1387,6 +1389,7 @@ export default class Player extends FakeEventTarget {
     if (this._engine) {
       this._engine.enableAdaptiveBitrate();
     }
+    this._playbackAttributesState.videoHeight = undefined;
```
SivanA-Kaltura, line 1392: why does it need to be reset ?

src/player.js
yairans: it's only for manual selection. on ABR mode `this._playbackAttributesState.videoHeight` is undefined

src/player.js
yairans: to not apply default track when ABR mode is active (auto-selection according the network)

src/player.js
OrenMe, line 2614: this should be just `videoTracks`

src/player.js
SivanA-Kaltura: what if we change video track, then change audio track, and then move to a new playlist item ?

src/player.js
OrenMe: I think the code is a bit too implicit in the way I read it, as it relies on the fact we have undefined value when we call `enableAdaptiveBitrate`.
Maybe better to put it in an if statement that relates on `isAdaptiveBitrateEnabled`.
```javascript
if (this.isAdaptiveBitrateEnabled()) {
  /*this is sort of like auto option for audio and text track*/
} else {
  /*handle manual selection*/
}
```

yairans: it has find in the end

src/player.js
OrenMe: I also think we should save the entire object of the track instead of just height, and handle the specific matching algorithm(if to match by width, height or label) afterwards in the `_setDefaultVideoTrack` method

src/player.js
yairans: `isAdaptiveBitrateEnabled` will be true after change media. this is the default.

src/player.js
yairans: how audio track selection related to video selection?

SivanA-Kaltura: oh, never mind, I thought the reset was also in selectTrack

src/player.js
yairans: done. first compare by label if not found then compare by height

src/player.js
```diff
@@ -2549,6 +2552,7 @@ export default class Player extends FakeEventTarget {
     let currentOrConfiguredAudioLang = this._playbackAttributesState.audioLanguage || playbackConfig.audioLanguage;
     this._setDefaultTrack<TextTrack>(this._getTextTracks(), currentOrConfiguredTextLang, offTextTrack);
     this._setDefaultTrack<AudioTrack>(this._getAudioTracks(), currentOrConfiguredAudioLang, activeTracks.audio);
+    this._setDefaultVideoTrack();
```
OrenMe, line 2555: the issue is talking about playlist but this will also now work for non playlist cases - is this intended?
I think we should put this behind a flag so it can be controlled, and the flag should be generic like `preserveVideoSelection` and by default it will be `false` and when loading playlist, internally in kaltura-player we will set it to true.
this will help us avoid breaking changes as well.
I also think we should think of preserving ABR vs manual bitrates as a feature so we can also use it for ABR, so maybe define `preserveVideoSelection` as an enum with options:
* `NONE`(as we have today) 
* `MANUAL`(the current feature in this PR) 
* `ABR`(the new proposed one)
* `ALL`(both manual and ABR)

yairans: as discussed with Ron L this is for manual selection only. and for any change media flow (not playlist only). it's very similar to captions selection or audio selection that we keep while the session

src/player.js
```diff
@@ -2601,6 +2605,20 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  /**
+   * Sets the video track selected by the user.
+   * @returns {void}
+   * @private
+   */
+  _setDefaultVideoTrack() {
```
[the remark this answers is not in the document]
yairans: added

src/player.js
```diff
@@ -2601,6 +2605,26 @@ export default class Player extends FakeEventTarget {
...
+    const sortedVideoTracks = this._getVideoTracks().sort((track1: VideoTrack, track2: VideoTrack) => track2.bandwidth - track1.bandwidth);
+    let selectedVideoTrack = sortedVideoTracks.find(
+      (track: VideoTrack) => track.label && track.label === this._playbackAttributesState.videoTrack?.label
```
JonathanTGold, line 2616: `track.label &&` is not redundant ?, anyway you can use the Optional chaining operator like you did later

src/player.js
```diff
@@ -2601,6 +2605,26 @@ export default class Player extends FakeEventTarget {
...
+   * @private
+   */
+  _setDefaultVideoTrack(): void {
+    const sortedVideoTracks = this._getVideoTracks().sort((track1: VideoTrack, track2: VideoTrack) => track2.bandwidth - track1.bandwidth);
+    let selectedVideoTrack = sortedVideoTracks.find(
+      (track: VideoTrack) => track.label && track.label === this._playbackAttributesState.videoTrack?.label
+    );
+    if (!selectedVideoTrack) {
+      selectedVideoTrack = sortedVideoTracks.find(
+        (track: VideoTrack) => track.height && track.height === this._playbackAttributesState.videoTrack?.height
```
JonathanTGold, line 2620: same as above

src/player.js
```diff
@@ -2601,6 +2605,26 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  /**
+   * Sets the video track selected by the user.
+   * @returns {void}
+   * @private
+   */
+  _setDefaultVideoTrack(): void {
+    const sortedVideoTracks = this._getVideoTracks().sort((track1: VideoTrack, track2: VideoTrack) => track2.bandwidth - track1.bandwidth);
```
JonathanTGold, line 2614: why the sort() is need before the find() ?

src/player.js
yairans: to ignore the case (mostly in tests) when the label/height is undefined so it equal `this._playbackAttributesState.videoTrack?.label`

src/player.js
yairans: when there are some tracks with the same label/height we want to keep the higher bandwidth, like we do in the [UI](https://github.com/kaltura/playkit-js-ui/blob/9d8a69af54817300355897dbef9d6db1215a67d7/src/components/settings/settings.js#L311)

SivanA-Kaltura (approved):

PR #645, opened 2022-03-13
fix(FEC-11904): Player_Web_Mac - The spinner appears when unmute the show

yairans:

Safari fires WAITING event once unmuting and immediately sends PLAYING
so prevent sending waiting event for too short buffering

Solves FEC-11904

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -1156,7 +1156,6 @@ export default class Html5 extends FakeEventTarget implements IEngine {
       if (!playing) {
         this.dispatchEvent(new FakeEvent(Html5EventType.WAITING));
       }
-      playing = false;
     }, SHORT_BUFFERING_TIMEOUT);
   }
 
--- a/src/engines/html5/html5.js
+++ b/src/engines/html5/html5.js
@@ -17,6 +17,8 @@ import getLogger from '../../utils/logger';
 import {DroppedFramesWatcher} from '../dropped-frames-watcher';
 import {ThumbnailInfo} from '../../thumbnail/thumbnail-info';
 
+const SHORT_BUFFERING_TIMEOUT: number = 200;
+
 /**
  * Html5 engine for playback.
  * @classdesc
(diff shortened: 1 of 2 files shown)


test/src/engines/html5/html5.spec.js
```diff
@@ -31,6 +32,20 @@ describe('Html5', () => {
     engine._el.should.be.a('HTMLVideoElement');
   });
 
+  it('should prevent too short buffering event', done => {
+    const progressiveSource = sourcesConfig.Mp4.progressive[0];
+    const engine = Html5.createEngine(progressiveSource);
+    engine._eventManager.listen(engine, Html5EventType.WAITING, () => {
+      done();
```
SivanA-Kaltura, line 39: so wont the test finish after the first WAITING event ?

src/engines/html5/html5.js
```diff
@@ -1145,6 +1148,18 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     }
   }
 
+  _handleWaiting(): void {
+    let playing = false;
+    this._eventManager.listenOnce(this._el, Html5EventType.PLAYING, () => (playing = true));
+    setTimeout(() => {
+      // prevent sending waiting event for too short buffering
+      if (!playing) {
+        this.dispatchEvent(new FakeEvent(Html5EventType.WAITING));
```
SivanA-Kaltura, line 1157: so if it's not playing WAITING events will keep being fired?

test/src/engines/html5/html5.spec.js
yairans: w/o the fix the test will get multiple done calls

src/engines/html5/html5.js
yairans: as it should

src/engines/html5/html5.js
```diff
@@ -1145,6 +1148,18 @@ export default class Html5 extends FakeEventTarget implements IEngine {
     }
   }
 
+  _handleWaiting(): void {
+    let playing = false;
+    this._eventManager.listenOnce(this._el, Html5EventType.PLAYING, () => (playing = true));
```
JonathanTGold, line 1153: Maybe [el.readyState](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState) can help  instead ?

src/engines/html5/html5.js
SivanA-Kaltura: do we need to fire the WAITING manually ? doesnt it fire anyway if the video is waiting ?

src/engines/html5/html5.js
```diff
@@ -1145,6 +1148,18 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+    setTimeout(() => {
+      // prevent sending waiting event for too short buffering
+      if (!playing) {
+        this.dispatchEvent(new FakeEvent(Html5EventType.WAITING));
+      }
+      playing = false;
```
JonathanTGold, line 1159: Isn’t redundant ? it will be reset each time `handleWaiting()` is called

test/src/engines/html5/html5.spec.js
SivanA-Kaltura: I'm not sure I understand what is being tested - don't we want to check the number of WAITING events ?

yairans: should be only one waiting because the shorter one should be ignored

src/engines/html5/html5.js
yairans: yes we do. the HTML5 class warps the video element and delegates its event outside see https://github.com/kaltura/playkit-js/blob/80afd171528293b1a310f0ac5cc1a5b216dfa5fa/src/engines/html5/html5.js#L302

src/engines/html5/html5.js
yairans: el.readyState is 2 once waiting in both real buffering and unmuting so it won't help

JonathanTGold (approved):

PR #658, opened 2022-07-31
fix(FEC-12431): [WEB] - [O2CZ] Player_Web_MAC_Safari - The progress bar jumps front and back when seek

SivanA-Kaltura:

Handle a scenario of this._liveEdge (in native-adapter) being 0 in a live stream, which caused isOnLiveEdge to be true when it shouldn't be.

Resolves FEC-12431.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -1213,7 +1213,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
       }
     } else {
       const liveEdge = this._getLiveEdge();
-      if (this._liveEdge !== liveEdge) {
+      if (this._liveEdge && this._liveEdge !== liveEdge) {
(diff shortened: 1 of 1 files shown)


giladna (approved):

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -1213,7 +1213,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
       }
     } else {
       const liveEdge = this._getLiveEdge();
-      if (this._liveEdge !== liveEdge) {
```
JonathanTGold, line 1216: better to change the way we are calculating the segment duration and calculate it only once according to the manifest like it does in shake player for example 
but this is a refactor task for the next sprint
iv open an issue see [FEC-12438](https://kaltura.atlassian.net/browse/FEC-12438)

JonathanTGold (approved):

PR #673, opened 2022-11-07
fix(FEC-12679): ISO Code for Chinese and Spanish in Multi Audio

Tzipi-kaltura:

**the issue:**
when the language track is 639-2 iso code and it's not the 639-1 + additional latter.
(for example: Spanish, 639-1: "es" 639-2: "spa")
the player doesn't recognize the language and it doesn't do the match between codes.

**the solution:**
add fields of extensionAudioLanguage and extensionTextLanguage to the configuration
in order to add additional checking in case the 639-2 iso code is totally different from the 639-1 iso code.

solves FEC-12679

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/flow-typed/types/playback-config.js
+++ b/flow-typed/types/playback-config.js
@@ -2,6 +2,8 @@
 declare type PKPlaybackConfigObject = {
   audioLanguage: string,
   textLanguage: string,
+  extensionAudioLanguage: string,
+  extensionTextLanguage: string,
   volume: number,
   playsinline: boolean,
   crossOrigin: string,
--- a/flow-typed/types/playback-config.js
+++ b/flow-typed/types/playback-config.js
@@ -2,8 +2,8 @@
 declare type PKPlaybackConfigObject = {
   audioLanguage: string,
   textLanguage: string,
-  extensionAudioLanguage: string,
-  extensionTextLanguage: string,
+  additionalAudioLanguage: string,
+  additionalTextLanguage: string,
   volume: number,
   playsinline: boolean,
   crossOrigin: string,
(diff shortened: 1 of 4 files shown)


flow-typed/types/playback-config.js
```diff
@@ -2,6 +2,8 @@
 declare type PKPlaybackConfigObject = {
   audioLanguage: string,
   textLanguage: string,
+  extensionAudioLanguage: string,
```
SivanA-Kaltura, line 5: I would use "secondaryXLanguage" or "additionalXLanguage"
"extensionXLanguage" is not clear enough

src/track/track.js
```diff
@@ -9,15 +9,21 @@ export default class Track {
...
+  static langComparer(inputLang: string, trackLang: string, extensionLanguage: ?string, equal: ?boolean): boolean {
     try {
       inputLang = inputLang.toLowerCase();
       trackLang = trackLang.toLowerCase();
+      extensionLanguage ? (extensionLanguage = extensionLanguage.toLowerCase()) : extensionLanguage;
       if (equal) {
-        return inputLang ? inputLang === trackLang : false;
+        if (inputLang && inputLang === trackLang) {
+          return true;
+        } else if (extensionLanguage && extensionLanguage === trackLang) {
```
SivanA-Kaltura, line 24: why is the && needed ? 
if it's undefined it won't be equal to trackLang

SivanA-Kaltura (changes requested):

SivanA-Kaltura:
@Tzipi-kaltura note the failing unit test in track.spec.js
you should add more unit tests for the new behavior as well

flow-typed/types/playback-config.js
Tzipi-kaltura: done

giladna (approved):

PR #687, opened 2022-12-20
fix(FEC-12122):[WEB][Youtube media][iOS][Safari][iPhone] - Full screen UI is not responsive for clicks

Tzipi-kaltura:

**the issue:**
full screen button is not responsive for clicks in youtube media on ios devices.

**the root cause:**
in case of ios deivces the full screen is the native full screen. 
the webkitEnterFullScreen function works only on the video element.
we tried to call the function on the element that wrapped the youtube iframe.

**the solution:**
in case of youtube media on ios devices we always call the enterInBrowserFullscreen function since we can't call the webkitEnterFullScreen. 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/fullscreen/fullscreen-controller.js
+++ b/src/fullscreen/fullscreen-controller.js
@@ -4,6 +4,7 @@ import Player from '../player';
 import FakeEvent from '../event/fake-event';
 import * as Utils from '../utils/util';
 import {ScreenOrientationType} from '../enums/screen-orientation-type';
+import {EngineType} from '../engines/engine-type';
 
 /**
  * The IOS fullscreen class name.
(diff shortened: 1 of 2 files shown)


src/fullscreen/fullscreen-controller.js
```diff
@@ -100,7 +100,7 @@ class FullscreenController {
         fullScreenElement = this._player.getView();
       }
       if (this._player.env.os.name === 'iOS') {
-        if (playbackConfig.inBrowserFullscreen && playbackConfig.playsinline) {
+        if ((playbackConfig.inBrowserFullscreen && playbackConfig.playsinline) || this._player._engineType === 'youtube') {
```
giladna, line 103: @Tzipi-kaltura  can we use constant instead of hard coded?

Tzipi-kaltura: @giladna , done.

giladna (approved):

PR #698, opened 2023-01-18
fix(FEC-12686): Web - When changing the caption from RUS to JAP in the middle of the video, all of the past caption are shown at once

SivanA-Kaltura:

- When adding active cues - advance the cue index, but don't add active cues if they are are expired
- When removing active cues - use filter instead of splicing in a loop, which caused some cues to not be removed.
- Reset cue index on text track selection to force recalculation of the cue index

Resolves FEC-12686

### CheckLists

- [ ] Docs have been updated


lianbenjamin (approved):
LGTM

PR #717, opened 2023-06-14
fix(FEC-13227): video starts unmuted in playlist after playing YT entry

lianbenjamin:

- bug fix

**the issue:**
when playing a playlist and the audio should be muted- after playing a YT entry, the player is becoming unmuted for the next entry.

**root cause:**
it seems that the playkit is checking the YT engine's `muted` value, while the youtube engine hasn't even loaded yet. Because of that, the autoplay handler is executing `onFallbackMutedAutoPlay` function instead of `onMutedAutoPlay` function. As a result, when playing a different entry after playing the YT one, the UI is unmuting the player- because of the `fallbackMutedAutoPlay` state.

**solution:**
at the point when we're checking the `muted` value, it is possible that the engine hasn't loaded yet (like in our case), so the muted value that we get is not reflecting the actual state of muted; the engine is returning false by default.
to handle this case, we need to check the muted value also in `_playbackAttributesState` and in the config itself, prior to checking the engine.

Solves FEC-13227

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -2122,11 +2122,11 @@ export default class Player extends FakeEventTarget {
 
     const isMuted = () => {
       Player._logger.debug('Checking muted value');
-      // at this point it is possible that the engine hasn't loaded yet, so this.muted is not reflecting the actual state.
+      // at this point it is possible that the engine hasn't loaded yet (like youtube), so this.muted is not reflecting the actual state.
       // first, check if someone set a value in muted; if not, then check the config; lastly, check the engine.
       const muted =
         (typeof this._playbackAttributesState.muted === 'boolean' && this._playbackAttributesState.muted) ||
-        this._config.playback.muted ||
+        (typeof this._config.playback.muted === 'boolean' && this._config.playback.muted) ||
         this.muted;
       Player._logger.debug('Muted value is:', muted);
       return muted;
--- a/src/player.js
+++ b/src/player.js
@@ -2122,13 +2122,13 @@ export default class Player extends FakeEventTarget {
 
     const isMuted = () => {
       Player._logger.debug('Checking muted value');
-      // at this point it is possible that the engine hasn't loaded yet, so this.muted is not refelcting the actual state.
+      // at this point it is possible that the engine hasn't loaded yet, so this.muted is not reflecting the actual state.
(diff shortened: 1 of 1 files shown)


src/player.js
```diff
@@ -1963,8 +1963,8 @@ export default class Player extends FakeEventTarget {
       this._eventManager.listen(this, Html5EventType.PAUSE, this._onPause.bind(this));
       this._eventManager.listen(this, Html5EventType.PLAYING, this._onPlaying.bind(this));
       this._eventManager.listen(this, Html5EventType.ENDED, this._onEnded.bind(this));
-      this._eventManager.listen(this, CustomEventType.MUTE_CHANGE, () => {
-        this._playbackAttributesState.muted = this.muted;
+      this._eventManager.listen(this, CustomEventType.MUTE_CHANGE, (event: FakeEvent) => {
+        this._playbackAttributesState.muted = event.payload?.mute || this.muted;
```
SivanA-Kaltura, line 1967: need to check if there wasn't some usecase that counted on the previous behavior

src/player.js
```diff
@@ -2120,6 +2120,18 @@ export default class Player extends FakeEventTarget {
       }
     });
 
+    const isMuted = () => {
+      Player._logger.debug('Checking muted value');
+      // at this point it is possible that the engine hasn't loaded yet, so this.muted is not reflecting the actual state.
+      // first, check if someone set a value in muted; if not, then check the config; lastly, check the engine.
+      const muted =
+        (typeof this._playbackAttributesState.muted === 'boolean' && this._playbackAttributesState.muted) ||
+        this._config.playback.muted ||
```
SivanA-Kaltura, line 2129: this._config.playback.muted can also be undefined, and we want to distinguish between undefined and false

SivanA-Kaltura (changes requested):

src/player.js
lianbenjamin: I tested manually the mute/unmute functionality for playlist and single media- seems fine to me. can't think of a use-case where we'd want to ignore the payload if it exists.

SivanA-Kaltura (approved):

PR #727, opened 2023-11-08
feat(sup-39815):[LG WebOS] add support for webOS initial bitrate option

inbalvasserman:
issue:
request to manage ABR in native HLS/DASH playback. 
need to support:options.adaptiveStreaming.bps, options.adaptiveStreaming.bps.start

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -466,7 +466,6 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           source.setAttribute('type', mimetype);
 
           if (this._config.useMediaOptionAttribute) {
-            NativeAdapter._logger.debug('this._config.abrEwmaDefaultEstimate', this._config.abrEwmaDefaultEstimate);
             const options = {};
             options.option = {};
             if (this._config.abrEwmaDefaultEstimate) {
@@ -475,7 +474,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
                 start: this._config.abrEwmaDefaultEstimate
               };
             }
-            NativeAdapter._logger.debug('options', options);
+            NativeAdapter._logger.debug('Setting mediaOption -', options);
             const mediaOption = encodeURI(JSON.stringify(options));
             source.setAttribute('type', mimetype + ';mediaOption=' + mediaOption);
           }
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -459,9 +459,10 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     requestFilterPromise = requestFilterPromise || Promise.resolve(pkRequest);
     requestFilterPromise
       .then(updatedRequest => {
-        if (this._sourceObj && this._config.useSourceTag) {
+        if (this._config.useSourceTag) {
           const source = document.createElement('source');
-          const mimetype = this._sourceObj.mimetype.toLowerCase();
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -456,7 +459,30 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
       .then(updatedRequest => {
-        this._videoElement.src = updatedRequest.url;
+        if (this._config.useSourceTag) {
+          const source = document.createElement('source');
+          const mimetype = this._sourceObj.mimetype.toLowerCase();
+          source.setAttribute('src', updatedRequest.url);
+          source.setAttribute('type', mimetype);
+
+          if (this._config.useMediaOptionAttribute) {
+            NativeAdapter._logger.debug('this._config.abrEwmaDefaultEstimate', this._config.abrEwmaDefaultEstimate);
```
OrenMe, line 469: Can be removed

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -456,7 +459,30 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+            const options = {};
+            options.option = {};
+            if (this._config.abrEwmaDefaultEstimate) {
+              options.option.adaptiveStreaming = {};
+              options.option.adaptiveStreaming.bps = {
+                start: this._config.abrEwmaDefaultEstimate
+              };
+            }
+            NativeAdapter._logger.debug('options', options);
```
OrenMe, line 478:
```suggestion
            NativeAdapter._logger.debug('Setting mediaOption - ', options);
```

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -456,7 +459,30 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+            const mediaOption = encodeURI(JSON.stringify(options));
```
SivanA-Kaltura, line 479: what needs to be encoded here ? is it only the src ? then you can run encode on the src instead of on the entire object

SivanA-Kaltura (changes requested):

src/engines/html5/media-source/adapters/native-adapter.js
OrenMe: according to LG entire object

OrenMe (dismissed):
LGTM

SivanA-Kaltura (approved):

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -461,7 +461,7 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
       .then(updatedRequest => {
         if (this._config.useSourceTag) {
           const source = document.createElement('source');
-          const mimetype = this._sourceObj.mimetype.toLowerCase();
+          const mimetype = this._sourceObj ? this._sourceObj.mimetype.toLowerCase() : 'application/dash+xml';
```
OrenMe, line 464: we can't put arbitrary value here, it must be coming from the source, if it is HLS but no mimetype is provided in config it might cause an error if incorrect one is provided

inbalvasserman: since there is an option that sourceobj is empty what do you suggest to put? because the original change cause errors in the test because of it.

OrenMe: U need to test it, I guess if there is no mimetype you should avoid setting the type all together
what are the errors if you set null? and are they in webos or also on web?

SivanA-Kaltura (approved):

PR #736, opened 2023-11-28
feat: support full webOS mediaOption object

OrenMe:

if useMediaOptionAttribute is true then first take all mediaOptionAttribute which is an object that passes as is the webLG options as defined in https://webostv.developer.lge.com/develop/guides/mediaoption-parameter If abrEwmaDefaultEstimate is defined so it passes it as override on top of the mediaOption in the 'option.adaptiveStreaming.bps' object

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.js
+++ b/src/engines/html5/media-source/adapters/native-adapter.js
@@ -467,13 +467,25 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           source.setAttribute('type', mimetype);
 
           if (this._config.useMediaOptionAttribute) {
-            const options = {};
-            options.option = {};
+            let options = {};
+            // https://webostv.developer.lge.com/develop/guides/mediaoption-parameter
+            if (this._config.mediaOptionAttribute) {
+              /**
+               * Undocumented option.
+               * Usage example:
+               * var video = document.querySelector('video');
+               * video.addEventListener("umsmediainfo", function(e) {
+               *     console.log(JSON.parse(e.detail));
+               * });
+               * {
+               *   useUMSMediaInfo: true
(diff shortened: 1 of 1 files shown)


src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -467,13 +467,25 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
           source.setAttribute('type', mimetype);
 
           if (this._config.useMediaOptionAttribute) {
-            const options = {};
-            options.option = {};
+            let options = {};
+            // https://webostv.developer.lge.com/develop/guides/mediaoption-parameter
```
MosheMaorKaltura, line 471: what is the purpose of posting this URL here?

src/engines/html5/media-source/adapters/native-adapter.js
```diff
@@ -467,13 +467,25 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
...
+            if (this._config.mediaOptionAttribute) {
+              /**
```
MosheMaorKaltura, line 473: remove obsolete code

src/engines/html5/media-source/adapters/native-adapter.js
OrenMe: this is the definition of possible configs

src/engines/html5/media-source/adapters/native-adapter.js
OrenMe: this is also documentation for possible usage of the config, I think this is good to keep as it is undocumented

MosheMaorKaltura (approved):

PR #787, opened 2024-06-23
fix(ADA-6): v7 player (accessibility) - Frames for Video Players Aren't Labeled

Tzipi-kaltura:

Please add a detailed description of the change, whether it's an enhancement or a bugfix.

**Issue:**
Iframe elements are labeled with uiConf/xxxxxx (the last parameter in the src url) in iframe lists (on screen readers) and not name.

**Solution:**
Add title element with the entry name

Solves [ADA-6](https://kaltura.atlassian.net/browse/ADA-6)

- [ ] Docs have been updated

[ADA-6]: https://kaltura.atlassian.net/browse/ADA-6?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -2020,7 +2020,7 @@ export default class Player extends FakeEventTarget {
         title = Utils.Dom.createElement('title');
         head.appendChild(title);
       }
-      title.innerHTML = this._sources.metadata.name;
+      title.innerHTML = this._sources.metadata.name ? this._sources.metadata.name : '';
     }
   }
 
--- a/src/player.ts
+++ b/src/player.ts
@@ -2013,13 +2013,15 @@ export default class Player extends FakeEventTarget {
    * @private
    */
   private _addTitleOnIframe(): void {
-    const head = Utils.Dom.getElementBySelector('head');
-    let title = head.querySelector('title');
-    if (!title){
-      title = Utils.Dom.createElement('title');
-      head.appendChild(title);
+    if (window.self !== window.top) {
+      const head = Utils.Dom.getElementBySelector('head');
+      let title = head.querySelector('title');
+      if (!title){
+        title = Utils.Dom.createElement('title');
+        head.appendChild(title);
+      }
+      title.innerHTML = this._sources.metadata.name;
     }
-    title.innerHTML = this._sources.metadata.name;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/player.ts
```diff
@@ -2006,6 +2007,21 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  /**
+   * In iframe embed add title element with the entry name
+   * @returns {void}
+   * @private
+   */
+  private _addTitleOnIframe(): void {
+    const head = Utils.Dom.getElementBySelector('head');
```
SivanA-Kaltura, line 2016: how will this match only iframe embeds ?
what if there is no name in metadata ?

SivanA-Kaltura (changes requested):

src/player.ts
Tzipi-kaltura: add the iframe check.
if there is no name title will be empty (per Sagi)

src/player.ts
```diff
@@ -2006,6 +2007,23 @@ export default class Player extends FakeEventTarget {
     }
   }
 
+  /**
+   * In iframe embed add title element with the entry name
+   * @returns {void}
+   * @private
+   */
+  private _addTitleOnIframe(): void {
+    if (window.self !== window.top) {
```
SivanA-Kaltura, line 2016: looking at answers from the web, it looks like both window.top and window.frameElement can throw an error in case of being embedded in a cross domain parent page

maybe using referrer would work better

SivanA-Kaltura (changes requested):

src/player.ts
```diff
@@ -2006,6 +2007,23 @@ export default class Player extends FakeEventTarget {
...
+      const head = Utils.Dom.getElementBySelector('head');
+      let title = head.querySelector('title');
+      if (!title){
+        title = Utils.Dom.createElement('title');
+        head.appendChild(title);
+      }
+      title.innerHTML = this._sources.metadata.name ? this._sources.metadata.name : '';
```
SivanA-Kaltura, line 2023: what if metadata is not set ?

SivanA-Kaltura (changes requested):

src/player.ts
SivanA-Kaltura: what if the title is not empty ? meaning, if the page containing the player already set an iframe text ?

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #788, opened 2024-07-04
fix(SUP-43106): bring captions in front of entry title/description

Lkaltura:

Please add a detailed description of the change, whether it's an enhancement or a bugfix.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -44,6 +44,7 @@
 }
 
 .playkit-subtitles {
+  z-index: 10;
   position: absolute;
   top: 0;
   bottom: 0;


Lkaltura:
issue:
when playing an audio entry that includes captions, the captions are hidden behind the title and description of the entry.
fix:
bring captions in front of the title/description 
resolves: SUP-43106

src/assets/style.css
```diff
@@ -44,6 +44,7 @@
 }
 
 .playkit-subtitles {
+  z-index: 10;
```
SivanA-Kaltura, line 47: how do subtitles look in non-audio entries after the change ? do they look ok ?

SivanA-Kaltura (changes requested):

src/assets/style.css
semarche-kaltura, line 47: I feel like the changes should be applied only for audio-entries. Also, need check how it behaves with dual-screen plugin that manage caption z-index as well  
https://github.com/kaltura/playkit-js-dual-screen/blob/e98dc19c8d19d8da36d5c0c9b2f07d29ad85612b/src/styles/global.scss#L1

semarche-kaltura (changes requested):

src/assets/style.css
Lkaltura: in non-audio entries, you don't have title/description on the media itself, so it's not the same case.
however, I did encounter this issue in both types:
image
in this case, shall we apply z-index to all the buttons ? if so, what should be taken into account to determine the value of it?

Lkaltura: this is how it would look like if : 
.player .bottom-bar { z-index = 3 }
and 
.playkit-subtitles { z-index = 2 }
image

let me know if something else needs to be considered here

Lkaltura: could you elaborate on the dual-screen ?
because there's title/descreption on the media like audio entries,
i assume you're talking about the child video. what is the expected behavior with regards to the caption? it's supposed to be behind the captions still no?

Lkaltura: either ways (z-index =1 or more) the captions would be in front of the 2 videos.

SivanA-Kaltura: ok, i agree with what sergey said below - we should localize the change to audio entries.
theres a variable on the global state called isAudio that should help with checking the entry type. see example [here](https://github.com/kaltura/playkit-js-ui/blob/91e5b697040240a35ddf8001d494326aac0abc36/src/components/audio-entry-details/audio-entry-details.tsx#L83)

SivanA-Kaltura (changes requested):

PR #795, opened 2024-09-15
fix(FEC-14080): Accessibility fix for the top bar and the bottom bar

SivanA-Kaltura:

Add video-filter div for use by ui when rendering bar gradients
Add class to indicate when native adapter displays native captions

Resolves FEC-14080

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/engines/html5/media-source/adapters/native-adapter.ts
+++ b/src/engines/html5/media-source/adapters/native-adapter.ts
@@ -220,6 +220,9 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     if (Utils.Object.hasPropertyPath(config, 'text.useNativeTextTrack')) {
       adapterConfig.displayTextTrack = Utils.Object.getPropertyPath(config, 'text.useNativeTextTrack');
     }
+    if (adapterConfig.displayTextTrack !== false) {
+      videoElement.classList.add('playkit-native-text');
+    }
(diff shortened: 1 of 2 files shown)


src/engines/html5/media-source/adapters/native-adapter.ts
```diff
@@ -220,6 +220,9 @@ export default class NativeAdapter extends BaseMediaSourceAdapter {
     if (Utils.Object.hasPropertyPath(config, 'text.useNativeTextTrack')) {
       adapterConfig.displayTextTrack = Utils.Object.getPropertyPath(config, 'text.useNativeTextTrack');
     }
+    if (adapterConfig.displayTextTrack !== false) {
+      videoElement.classList.add('playkit-native-text');
```
lianbenjamin, line 224: do we need to remove the class at some point?
is it possible that it will be added when it already exists?
Also, maybe better to export the class name to a const.

lianbenjamin (changes requested):

src/engines/html5/media-source/adapters/native-adapter.ts
SivanA-Kaltura: classList won't add it again, but yeah, i guess we want to remove it when the adapter is destroyed or reset

lianbenjamin (approved):

PR #797, opened 2024-11-05
feat(FEC-14087): define audio-description audio tracks

semarche-kaltura:

resolves: https://kaltura.atlassian.net/browse/FEC-14087

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -2470,6 +2470,27 @@ export default class Player extends FakeEventTarget {
    */
   public _updateTracks(tracks: Array<Track>): void {
     Player._logger.debug('Tracks changed', tracks);
+
+    if (this._sources.metadata?.audioFlavors) {
+      let audioTracksIndex = 0;
+      // iterate over the audio tracks and set the isAudioDescription flag based on the audioFlavors metadata
+      tracks.forEach((track) => {
+        if (track instanceof AudioTrack) {
+          track.isAudioDescription = Boolean(this._sources.metadata.audioFlavors?.[audioTracksIndex]?.isAudioDescription);
+          if (track.isAudioDescription) {
+            // set the language to ad-<language> for audio description tracks
+            // track.language = `ad-${track.language}`;
+            track.language = `${track.language}-ad`;
+            if (!this._sources.metadata?.audioFlavors?.[audioTracksIndex]?.label) {
+              // add "Audio Description" to the label if custom label is not provided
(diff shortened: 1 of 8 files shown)


src/player.ts
```diff
@@ -2470,6 +2470,26 @@ export default class Player extends FakeEventTarget {
    */
   public _updateTracks(tracks: Array<Track>): void {
     Player._logger.debug('Tracks changed', tracks);
+
+    if (this.config.playback.updateAudioDescriptionFlavors && this._sources.metadata?.audioFlavors) {
```
SivanA-Kaltura, line 2474: Can we have audioFlavors as an empty array ?
Maybe text audioFlavors?.length ?

src/player.ts
```diff
@@ -2470,6 +2470,26 @@ export default class Player extends FakeEventTarget {
...
+      let audioTracksIndex = 0;
+      // iterate over the audio tracks and set the isAudioDescription flag based on the audioFlavors metadata
+      tracks.forEach((track) => {
```
SivanA-Kaltura, line 2477: I think the most generic solution, to prepare for a future case of 
"if we have a track with metadata A, then do logic B"
would be to extract the logic handling of handling the track to a separate file,
and then just leave the condition here

so something like
if (this.config.playback.updateAudioDescriptionFlavors) {
   audioDescriptionTrackHandler(track);
}

we could also make it even more generic and add a factory, but that's less important if we only have one usecase like this so far

SivanA-Kaltura (changes requested):

src/track/audio-track.ts
```diff
@@ -1,10 +1,41 @@
...
+class AudioTrack extends Track {
+  /**
+   * @member - is audio asset has audio description tag
+   * @type {boolean|undefined}
+   */
+  public isAudioDescription?: boolean;
+}
+
+export function audioDescriptionTrackHandler(tracks: AudioTrack[], audioFlavors?: Array<any>): void {
+  if (tracks?.length) {
```
SivanA-Kaltura, line 21: if is not set or flavors.length is 0 i think we can just check it once at the top of the function and return tracks as is, instead of checking audioFlavors?[index]? every time

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #809, opened 2025-02-20
fix (SUP-46985): Replay button unresponsive when PiP is activated

yambenari24:
Issue:
On safari, when PiP is open, video ends and try to play it again with start over, nothing happened.
Safari PiP locks the video in an "ended" state, preventing .play() from restarting playback unless the video is explicitly paused and its currentTime is reset.
Unlike Chrome, Safari doesn't automatically reset currentTime to 0 when a video ends.
 
Fix:
Added a browser check to detect Safari using the user agent.
Applied a conditional reset of the video (pause() and currentTime = 0) only when:
The browser is Safari.
The video is in Picture-in-Picture (PiP) mode.

solved [SUP-46985](https://kaltura.atlassian.net/browse/SUP-46985)

[SUP-46985]: https://kaltura.atlassian.net/browse/SUP-46985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/engines/html5/html5.ts
+++ b/src/engines/html5/html5.ts
@@ -519,6 +519,14 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @returns {?Promise<*>} - play promise
    */
   public play(): Promise<void> {
+    const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+    if (isSafari) {
+      if (this._el.currentTime >= this._el.duration - 0.1 || this._el.ended) {
+        this._el.pause();
+        this._el.currentTime = 0;
+      }
+    }
+
     const playPromise = this._el.play();
     if (playPromise) {
       playPromise.catch((err) => this.dispatchEvent(new FakeEvent(CustomEventType.PLAY_FAILED, { error: err })));
--- a/src/engines/html5/html5.ts
+++ b/src/engines/html5/html5.ts
@@ -19,7 +19,7 @@ import { CapabilityResult, ICapability } from '../../types';
 import { PKABRRestrictionObject, PKDrmConfigObject, PKDrmDataObject, PKMediaSourceObject, PKVideoElementStore } from '../../types';
 import { IEngine } from '../../types';
 import Track from '../../track/track';
-import { Env } from '../../../src/playkit';
+import Env from '../../../src/utils/env';
 
 const SHORT_BUFFERING_TIMEOUT: number = 200;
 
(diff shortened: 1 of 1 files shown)


src/engines/html5/html5.ts
```diff
@@ -519,6 +519,14 @@ export default class Html5 extends FakeEventTarget implements IEngine {
    * @returns {?Promise<*>} - play promise
    */
   public play(): Promise<void> {
+    const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
```
SivanA-Kaltura, line 522: you can use isSafari

src/engines/html5/html5.ts
```diff
@@ -519,6 +519,14 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+    if (isSafari && this.isInPictureInPicture) {
+      if (this._el.currentTime >= this._el.duration - 0.1 || this._el.ended) {
```
SivanA-Kaltura, line 524: why 0.1 ?

SivanA-Kaltura (changes requested):

src/engines/html5/html5.ts
```diff
@@ -519,6 +519,14 @@ export default class Html5 extends FakeEventTarget implements IEngine {
...
+        this._el.pause();
+        this._el.currentTime = 0;
```
SivanA-Kaltura, line 526: live video doesn't necessarily start from 0

SivanA-Kaltura (changes requested):

src/engines/html5/html5.ts
yambenari24: the check this._el.currentTime >= this._el.duration might fail due to floating-point precision errors. CurrentTime and duration are floating-point numbers, and sometimes currentTime may never exactly equal duration due to tiny decimal inaccuracies. By subtracting 0.1, you're introducing a tolerance, allowing the condition to pass even if currentTime is slightly below duration due to those precision issues.
Do you have a better solution?

SivanA-Kaltura: round up the number ? multiply by 1000 and round ?
it's just that 100ms is a noticeable difference

src/engines/html5/html5.ts
yambenari24: If live ends, and there is not VOD recording, then it's not allowed to do start over.
If there VOD recording exists, we can play over, but it's won't be live anymore.
Therefore, add the condition of !live can solve it.

SivanA-Kaltura (approved):

PR #824, opened 2025-07-24
Add prioritizeAudioDescription flag in playback config for default sorting audio tracks in the same language

roeedean:

Adds `prioritizeAudioDescription` flag in playback config:
When set to true, audio description tracks (identified by language prefix `"ad-"`) will be prioritized and appear at the top of the audio tracks list. Otherwise, regular audio tracks appear first.


src/player.ts
```diff
@@ -2596,7 +2596,37 @@ export default class Player extends FakeEventTarget {
...
+          return prioritizeTracks
+            ? (aIsAudioDescription ? -1 : 1)  // Audio description tracks first
+            : (aIsAudioDescription ? 1 : -1); // Regular tracks first
+        }
+
+        // Order alphabetically by label (A-Z) only if alphabeticalSort has a boolean value
+        if (typeof alphabeticalSort === 'boolean') {
+          const aLabel = a.label?.toLowerCase() || a.language || '';
+          const bLabel = b.label?.toLowerCase() || b.language || '';
+          if (aLabel < bLabel) return -1;
```
SivanA-Kaltura, line 2621: what about using aLabel.localeCompare(bLabel) ?

src/player.ts
```diff
@@ -2596,7 +2596,37 @@ export default class Player extends FakeEventTarget {
...
+    const audioTracks = this._getTracksByType<AudioTrack>(AudioTrack);
+
+    const prioritizeTracks = this._config.playback.prioritizeAudioDescription;
+    const alphabeticalSort = this._config.playback.alphabeticalSort;
+
+    // Only apply sorting if prioritizeTracks or alphabeticalSort have boolean values
+    if (typeof prioritizeTracks === 'boolean' || typeof alphabeticalSort === 'boolean') {
+      return audioTracks.sort((a, b) => {
+        const aIsAudioDescription = a.language?.startsWith('ad-') || false;
+        const bIsAudioDescription = b.language?.startsWith('ad-') || false;
```
SivanA-Kaltura, line 2608: why not just b.language?.startsWith('ad-') ?

src/player.ts
```diff
@@ -2596,7 +2596,37 @@ export default class Player extends FakeEventTarget {
...
   private _getAudioTracks(): Array<AudioTrack> {
-    return this._getTracksByType<AudioTrack>(AudioTrack);
+    const audioTracks = this._getTracksByType<AudioTrack>(AudioTrack);
+
+    const prioritizeTracks = this._config.playback.prioritizeAudioDescription;
+    const alphabeticalSort = this._config.playback.alphabeticalSort;
+
+    // Only apply sorting if prioritizeTracks or alphabeticalSort have boolean values
+    if (typeof prioritizeTracks === 'boolean' || typeof alphabeticalSort === 'boolean') {
+      return audioTracks.sort((a, b) => {
```
SivanA-Kaltura, line 2606: i think it would be more readable to create 3 functions, one per sort type, and pass one of them to sort() by condition
there's no point in testing the condition inside the loop

so something like

  private _getAudioTracks(): Array<AudioTrack> {
    const audioTracks = this._getTracksByType<AudioTrack>(AudioTrack);

    const prioritizeTracks = this._config.playback.prioritizeAudioDescription;
    const alphabeticalSort = this._config.playback.alphabeticalSort;

    function sortAudioDescriptionFirst(a, b) {
        const aIsAudioDescription = a.language?.startsWith('ad-');
        const bIsAudioDescription = b.language?.startsWith('ad-');

        return aIsAudioDescription !== bIsAudioDescription ? 1 : -1;
    }

    function sortNormalTracksFirst(a, b) {
        return -1 * sortAudioDescriptionFirst(a, b);
    }

    function sortAlphabetically(a, b) {
        const aLabel = a.label?.toLowerCase() || a.language || '';
        const bLabel = b.label?.toLowerCase() || b.language || '';

        return aLabel.localeCompare(bLabel);
    }

    if (alphabeticalSort) {
        return audioTracks.sort(sortAlphabetically);
    } else if (prioritizeTracks === true) {
        return audioTracks.sort(sortAudioDescriptionFirst);
    } else if (prioritizeTracks === false) {
        return audioTracks.sort(sortNormalTracksFirst);
    }

    return audioTracks;
  }

SivanA-Kaltura (changes requested):

src/player.ts
roeedean: Fixed

src/player.ts
roeedean: Fixed with `!!b.language?.startsWith('ad-')`

SivanA-Kaltura (approved):

PR #879, opened 2026-05-15
fix(audio-tracks): use Intl.DisplayNames for native language labels (SUP-52289)

MosheMaorKaltura:
## Problem

On iOS Safari, when the player falls back from DASH to HLS, audio track labels are sourced from the HLS manifest NAME= attribute (English words: "Spanish", "Japanese"). The DASH code path derives labels from the ISO language code which the player translates. This causes inconsistent audio track labels depending on the streaming format.

## Root Cause

native-adapter.ts _getParsedAudioTracks sets label: audioTracks[i].label directly from the browser's AudioTrack.label property, which is populated from the HLS manifest NAME= field. There is no normalisation to native script name.

## Fix

Added getNativeLanguageName(langCode, fallback) in src/utils/locale.ts using Intl.DisplayNames to convert an ISO 639 code into the native script name (e.g. "es" -> "español", "ja" -> "日本語"). The function is exported from playkit.ts so downstream adapters can import it.

Applied the helper in native-adapter.ts _getParsedAudioTracks:

  label: getNativeLanguageName(audioTracks[i].language, audioTracks[i].label),

The ISO language code is the primary input; the original manifest label is the safe fallback.

## Files Changed

- src/utils/locale.ts — added getNativeLanguageName export
- src/playkit.ts — re-exported getNativeLanguageName
- src/engines/html5/media-source/adapters/native-adapter.ts — applied helper at label assignment
- tests/e2e/utils/get-native-language-name.spec.js — regression tests (6 cases, all PASS)

## Tests

- Regression test: tests/e2e/utils/get-native-language-name.spec.js — PASS
- Full suite: 509 tests completed, 0 failures

## Companion PR

kaltura/playkit-js-hls — imports getNativeLanguageName and applies it in _parseAudioTracks

## Jira

https://kaltura.atlassian.net/browse/SUP-52289

### DIFF
--- a/api-extractor/playkit-js.api.json
+++ b/api-extractor/playkit-js.api.json
@@ -288,12 +288,174 @@
           "isAbstract": false,
           "name": "AudioTrack",
           "preserveMemberOrder": false,
-          "members": [],
+          "members": [
+            {
+              "kind": "Constructor",
+              "canonicalReference": "@playkit-js/playkit-js!AudioTrack:constructor(1)",
+              "docComment": "/**\n * @constructor Constructs a new instance of the `AudioTrack` class\n *\n * @param settings - The track settings object.\n */\n",
+              "excerptTokens": [
+                {
+                  "kind": "Content",
+                  "text": "constructor(settings?: "
+                },
+                {
+                  "kind": "Content",
+                  "text": "any"
+                },
+                {
+                  "kind": "Content",
+                  "text": ");"
+                }
+              ],
+              "releaseTag": "Public",
+              "isProtected": false,
+              "overloadIndex": 1,
+              "parameters": [
+                {
+                  "parameterName": "settings",
+                  "parameterTypeTokenRange": {
+                    "startIndex": 1,
+                    "endIndex": 2
+                  },
+                  "isOptional": true
+                }
+              ]
+            },
+            {
+              "kind": "Property",
+              "canonicalReference": "@playkit-js/playkit-js!AudioTrack#flavorId:member",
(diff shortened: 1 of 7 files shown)


MosheMaorKaltura:
## Code Review Remarks — Response

Addressed in commit 05e4cff:

**1. Browser Compatibility Comment ✅**
Added the comment above `_displayNamesCache` in `src/utils/locale.ts`:
```ts
// Intl.DisplayNames is supported in modern browsers (Chrome 81+, Firefox 86+, Safari 14.1+).
// Older browsers will fall back to the provided fallback value.
```

**2. Test File Location**
There is no `tests/unit/` directory in this project — the only spec directory is `tests/e2e/`. The `tests/utils/` directory contains test helper utilities (`test-utils.js`), not specs. Keeping the spec in `tests/e2e/utils/` is the correct location for this project's conventions. No change made.

**3. Case Sensitivity**
`Intl.DisplayNames.of()` returns the native name in its natural casing (e.g. `"Español"` with capital E, which is correct). The test uses `.toLowerCase()` only for the assertion to guard against potential casing differences across browser implementations — the actual value passed to the track label is the un-normalised native name, which is the expected UI output. No change needed.

**4. Memoization ✅**
Added a module-level `_displayNamesCache: Map<string, Intl.DisplayNames>` — each unique `langCode` gets one `Intl.DisplayNames` instance, reused on subsequent calls.

**5. API Extractor Changes**
`api-extractor/playkit-js.api.json` is already tracked on `main` (20k+ lines predating this PR). The 1242-line diff reflects the newly exported `getNativeLanguageName` function being documented. The file is auto-generated but committed intentionally as a public API snapshot — this is the existing pattern in the repo. No change made.

lianbenjamin (approved):

PR #888, opened 2026-07-14
fix(ADA-3093): Remove Talkback stop

valentinStoicaK:
This fixes https://kaltura.atlassian.net/browse/ADA-3093

The ticket's issue is multiple swipes when reaching pre-playback button while using android Talkback. Talkback sees the playkit-container and the playkit-poster as accessibility nodes, because of the use of tabindex="-1". Tabindex="-1" should be used to stop focusing an interactive item or to make it be focused programmatically. None of these cases are being used here, so removing it fixes one of the multiple swipes for Talkback

### CheckLists

- [ ] Docs have been updated


roeedean (commented):
@valentinStoicaK please add in the description why this change does exactly and why.

roeedean (approved):

